PR #22, opened 2014-08-01
Correcting invalid linker flag for clang on other platforms

### LINKED ISSUE #20: The -stdlib flag breaks compile on linux using clang
In CmakeLists.txt, using linux with clang 3.4.2, I have to change: 

```
set(CMAKE_CXX_FLAGS "-std=c++11 -stdlib=libc++")
```

to

```
set(CMAKE_CXX_FLAGS "-std=c++11")
```

in order for it to compile.   If I use the "-stdlib" flag, the compile can't find standard library headers (e.g., <list>).

For example:

```
[  1%] Building CXX object contrib/json11/CMakeFiles/json11.dir/json11.cpp.o
In file included from /home/ndoss/autowiring/autowiring/contrib/json11/json11.cpp:22:
/home/ndoss/autowiring/autowiring/contrib/json11/json11.hpp:53:10: fatal error: 
      'string' file not found
#include <string>
         ^
1 error generated.
```

codemercenary:
fixes #20


codemercenary:
#435, #268, #258, #208, #186.

So, so many pull requests that were aimed squarely at fixing libstdc support.  Gone, but not forgotten, I suppose.

PR #27, opened 2014-08-04
Filter exceptions thrown by AutoDesired

codemercenary:
I shouldn't have written AutoDesired to just blindly eat any exception that comes along.  I'm adding handling so that AutoDesired invokes all of the standard exception filters when an AutoDesired request fails.

I'm not really sure yet what AutoDesired is going to be; at this point, it's just an experimental proposition.  I'm open to suggestions on how it should behave--perhaps reworking the ExceptionFilter interface to have a concept of a nonfatal exception is a good idea?

### DIFF
--- a/autowiring/Autowired.h
+++ b/autowiring/Autowired.h
@@ -281,7 +281,9 @@ public:
 ///
 ///  try {
 ///    AutoCurrentContext()->Inject&lt;T&gt;();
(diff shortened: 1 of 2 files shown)


codemercenary:
What the heck even is `AutoDesired`?  I don't even remember anymore.

PR #28, opened 2014-08-04
Removed EventReceiver

gtremper:
- Types no longer need to inherit from `EventReceiver` to use Autowiring events functionality.
- Deprecated `deferred_ptr`, which wasn't used anywhere
- Renamed `seq` and `gen_seq` to `index_tuple` and `gen_index_tuple`


codemercenary:
This one was such a fiasco.  I can't believe I thought an `EventReceiver` common base type was even a good idea.

PR #32, opened 2014-08-05
Demangle type names on AutoNetServer

gtremper:
- Added `demangle()` function to AutoNetServer
- Turned off activity logging for websocketpp


codemercenary:
I'm glad that, in the end, we managed to get `autowiring::demangle` to behave consistently.  I remember all of the backflips you had to do to get the original implementation working properly because of its inconsistent returns.

PR #34, opened 2014-08-07
Header organization

gtremper:
Cleanup up a couple redundant parts in Autowiring. More specifically...
- Created `TypeIdentifier` that's shared by both TypeRegistry and EventRegistry
- Removed `CoreContext.h` include from TypeRegistry
  - Can include `TypeRegistry.h` without creating circular dependencies with `CoreContext`
  - Consolidated multiple TypeRegistry utterances in `Autowiring.h` to a single utterance in `CoreContext.h`
- Consolidated `autowiring::` and `Auto::` namespaces
- Consolidated 2 separate definitions of `index_tuple`

### DIFF
--- a/autowiring/CoreContext.h
+++ b/autowiring/CoreContext.h
@@ -938,6 +938,9 @@ public:
   }
 };
 
+/// <summary>
+/// Forward-declarable version of CoreContext::InjectCurrent
+/// </summary>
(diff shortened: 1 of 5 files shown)


codemercenary:
Almost forgot that we even had an `Auto::` namespace.

PR #35, opened 2014-08-07
Added Doxygen support

gtremper:
Install doxygen [here](www.doxygen.org/) or with `brew install doxygen`

Create doxygen documention in `docs/` with this command from the root project directory
`doxygen Doxyfile`


codemercenary:
Wow, for some reason I thought it was @jo3w4rd who enabled Doxygen.  Actually, you made the first commits to [autowiring.io](http://www.autowiring.io) and even registered the domain.

PR #314, opened 2015-01-01
Rename boost to autoboost

codemercenary:
Huge modification, requiring adjustements to namespaces, macros, file names, and header guards.  This change introduces a private version of Boost 1.57.0 that can be linked externally on platforms which do not have full STL11 support without colliding in any way with those platforms' independent use of boost.

Special thanks to @gittyupagain for getting the majority of this work started.  This PR has been squashed, though, in order to save on repo size.


codemercenary:
Thank you @gtremper

PR #316, opened 2015-01-06
Only install autoboost headers on platforms that need it

gtremper:


CMakeLists.txt
```diff
@@ -220,11 +220,13 @@ if(NOT AUTOWIRING_IS_EMBEDDED)
   )
 
   # Install autoboost headers
-  install(
-    DIRECTORY ${PROJECT_SOURCE_DIR}/contrib/autoboost/autoboost
-    DESTINATION include
-    COMPONENT autowiring
-  )
+  if(NOT autowiring_USE_LIBCXX)
```
jnguyen75: do you also need the "if(APPLE)" check as well?

codemercenary: No, because autowiring_USE_LIBCXX is presently set to "on" on all platforms, except on Mac, where it might be set to "off".

PR #330, opened 2015-01-14
Correct empty auto_prev satisfaction

gtremper:
`auto_prev`s are now correctly satisfied with an empty value when the previous packet can no longer satisfy the type.

Also, for each decoration for type `T` with timeshift `N`, decorations for type `T` and timeshifts `< N` are guaranteed to exist


gtremper:
Don't merge me yet

gtremper:
ready to merge

PR #331, opened 2015-01-15
Modify DispositionState to use an enum to describe it's level of satisfaction

tnitz:
The PartlySatisfied state is intended for use with the upcoming multi-decorate feature, to distinguish the state between Unsatisfied and Satisfied, where we know that not every provider of a decoration has given us one.

### DIFF
--- a/autowiring/AutoPacket.h
+++ b/autowiring/AutoPacket.h
@@ -256,7 +256,7 @@ public:
   bool Get(std::shared_ptr<const T>& out, int tshift = 0) const {
     std::lock_guard<std::mutex> lk(m_lock);
     auto deco = m_decorations.find(DecorationKey(typeid(auto_id<T>), tshift));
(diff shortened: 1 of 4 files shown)


gtremper:
looks good

codemercenary:
Testing on Windows

autowiring/AutoPacket.h
```diff
@@ -332,12 +332,12 @@ class AutoPacket:
       std::lock_guard<std::mutex> lk(m_lock);
       auto& entry = m_decorations[key];
       entry.SetKey(key); // Ensure correct type if instantiated here
-      if(entry.satisfied ||
-         entry.isCheckedOut)
+      if(entry.m_state == DispositionState::PartlySatisfied ||
```
codemercenary: I see why you need `PartlySatisfied`, now.  I guess there's not really any avoiding it, in order to prevent ambiguity.

PR #351, opened 2015-01-21
Feature multi in

tnitz:
I believe this is a good first phase of the multi-in paradigm. I would still like to add earlier failings when auto-filters which are in combination illegal are added to a packet factory, but this detection is not currently done in develop either.


tnitz:
@codemercenary 's additions to this branch look good to me. I believe this branch is finally ready to merge.

PR #353, opened 2015-01-23
Add support for delayed Initiate

codemercenary:
This change alters the way that startup occurs on child contexts whose parents are not yet started.  Formerly, invoking `CoreContext::Initiate` on a child context where `Initiate` was not called on the parent would result in recursive initiation of all parents up to the global context.  Now, the child will be transitioned into an initiated metastate where it is ready to run but not yet started.


tnitz:
Skipping this one for now due to a concern @gittyupagain expressed to me.

PR #357, opened 2015-01-24
Slight enhancement to the performance of auto_arg

codemercenary:
This enhancement eliminates a double lookup in cases where `std::shared_ptr` is accepted as an input type to an AutoFilter.  It also reduces the size of auto_arg by eliminating an unneeded member variable.


codemercenary:
Depends on #360

PR #380, opened 2015-01-29
Cleaner fix to "has no symbols" warning on Mac

codemercenary:
Dummy variables are used often to correct issues like this one.  At some point, however, we should consider either deprecating or making use of the uuid concept rather than just patching it like this.


gittyupagain:
Proposing to close and use PR #381 instead.

PR #388, opened 2015-02-03
Rename 'Object' to 'CoreObject'

gtremper:
The name `Object` is too generic and could clash with other libraries.


codemercenary:
Here's a good one to ensure we follow through with.  I think that eliminating the name `Object` is one of the last barriers to a good 1.0 release.

Issue #390, opened 2015-02-03
Provide TLS security support for AutoNetServer


jnguyen75:
As we are using AutoNetServer as the RPC messaging, we should look into adding TLS support so that AutoNetServer can communicate with web applications with a more secure protocol

pehrlich:
(Especially as some of our internal tooling demands https for proper security)

gtremper:
The transport layer has been separated from AutoNet, so an external secure websocket server can be used.

codemercenary:
Going to call this one "closed," in that case.

PR #396, opened 2015-02-04
Secure AutoNetServer

gtremper:
Add TLS version of `AutoNetServerImpl`. The insecure version of the server is the default. This will finish #390


codemercenary:
Almost forgot about this.  It's a good thing that we did it this way, otherwise we'd have an OpenSSL dependency to add to our list of problems.

PR #397, opened 2015-02-05
AutoNet trasport layer

gtremper:
Add separate transport layer for AutoNet. This closes #396


codemercenary:
lol wats a trasport layer

PR #410, opened 2015-02-11
Add virtual destructor to AutoNetTransport

gtremper:


codemercenary:
Thank you for still contributing to Autowiring even while you were on the tracking team.  Here's hoping that you continue to make contributions from time to time.

PR #416, opened 2015-02-16
Fix overflow problems with QueryPerformanceCounter() by using doubles

jhe-:
This is needed for PR #1376 in platform.

### DIFF
--- a/autowiring/C++11/chrono_with_profiling_clock.h
+++ b/autowiring/C++11/chrono_with_profiling_clock.h
@@ -27,11 +27,11 @@ extern "C" {
 }
 
 namespace {
-  const long long g_Frequency = []()
+  const double g_nanosecs_per_tic = []()
   {
     int64_t frequency;
     QueryPerformanceFrequency(&reinterpret_cast<LARGE_INTEGER&>(frequency));
-    return frequency;
+    return 1e9 / static_cast<double>(frequency);
   }();
 }
 
@@ -48,7 +48,7 @@ namespace std {
       static time_point now() {
         int64_t count;
         QueryPerformanceCounter(&reinterpret_cast<LARGE_INTEGER&>(count));
-        return time_point(duration(count * static_cast<rep>(period::den) / g_Frequency));
+        return time_point(duration(static_cast<int64_t>(static_cast<double>(count) * g_nanosecs_per_tic)));
       }
     };
   }


autowiring/C++11/chrono_with_profiling_clock.h
```diff
@@ -27,11 +27,11 @@ extern "C" {
 }
 
 namespace {
-  const long long g_Frequency = []()
+  const double g_nanosecs_per_tic = []()
```
codemercenary: Recommend we use `int64_t` instead.  There is no need to invoke the FPU here.

codemercenary:
I'm not completely convinced of the reasoning in this pull request.  What, exactly, was overflowing?

jhe-:
See my philosophical post :p http://sf-github.leap.corp/leapmotion/platform/pull/1370.

codemercenary:
I'll have a look, but recommend you completely justify here as well.  This is a public repository, after all.

jhe-:
Excerpted:


There are no severe problems inherent with QPC. Boost has always used QPC, and VS2015's std::chrono will use QPC. Yes, it has some bugs with multithreading and some hardware-specific defects, but for our timing purposes (μs for long intervals) it's the best we can get. The issues only arise from improper handling of its output. I dug a bit deeper and compared Walter's profiling_clock (which was based on a StackOverflow answer) to Boost's high_resolution_clock.

Walter's class:

    count * 1000000000L / frequency

Even though the final result fits into 64 bits, the value of (count * 1e9) can easily overflow, and is certainly what was causing the visualizer [bug](https://leapmotion.atlassian.net/browse/PLAT-3035). On a typical Windows box where the QPC frequency is 3.58 MHz, it only takes 2^63/(3.58e6 * 1e9) = 2576 seconds of system uptime before overflow happens. So all the time, basically.

[Boost avoids this by using doubles](https://github.com/ryppl/boost-history/blob/trunk/boost/chrono/detail/inlined/win/chrono.hpp#L31):

    (int64_t)(count * (double)(1000000000L / frequency))

(An even more accurate way would be:)

    (int64_t)(count * (1e9 / (double)frequency))

The dynamic range of doubles means that it will not overflow. The 52-bit mantissa means it won't have rounding errors either. Since neither of these problems are present, this value is accurate to the frequency of QPC itself.

So the solution is to [change Walter's class to use doubles](https://github.com/leapmotion/autowiring/pull/416), but [keep the use of profiling_clock](http://sf-github.leap.corp/leapmotion/platform/pull/1376). There's a million other ways you could do it. You could use a 128-bit int. But this works and is the simplest.

codemercenary:
"It works" isn't usually sufficient for me, but in this case I can't think of a better answer.  Let me think about this for a little while, and if I can't come up with a better idea, we can merge it.

Note that we won't be able to cut another release of autowiring in time for a release of Leap, we will need to make use of an alternative workaround to fix the problem in develop.

jhe-:
Ok, I'll just resurrect Utility/profiling_clock.h in the platform branch.

codemercenary:
Roger that.

jhe-:
(I hope one of your ideas isn't having some hybrid where you use a GetTickCount-based timer to correct for jitters in the QPC-based timer. That would get messy fast.)

codemercenary:
No, I wouldn't want to mix any counter types.

I was thinking about using the Windows [Multiply128](https://msdn.microsoft.com/en-us/library/windows/desktop/hh802931(v=vs.85).aspx) function to get what I want.  This code is all Windows-only, anyway, so it would technically do the trick.

jhe-:
Yeah, I was considering that. It would be a cleaner solution, I was just scared when I saw winnt.

codemercenary:
The function can be forward declared, leave it to the user to provide the necessary link libraries.

jhe-:
Hm, that's a good point.

jhe-:
There's no over-the-counter Divide128 though. :(

codemercenary:
A clever trick would be to find the modular inverse of `g_Frequency` and then multiply that by 1e9 statically.  The [extended euclidian algorithm](http://en.wikipedia.org/wiki/Extended_Euclidean_algorithm) is both fast and sufficiently simple for this purpose, sample code is near the bottom.  The only caveat is that `g_Frequency` must be relatively prime to 2^64--a quality easy enough to assure if we are OK with nudging its value by a tiny bit if it happens to have no other factors than 2.

jhe-:
If

    1e9*mod_inv(g_Frequency)

is the value you're calculating, I don't think it will work. One can't declare that 1e9*mod_inv(g_Frequency) is a substitute for 1e9/g_Frequency (the desired value).

One can only say that:

g_Frequency * (1e9*mod_inv(g_Frequency)) ≡ g_Frequency * desired_value     (mod 2^64)

which means that you would need to carry around g_Frequency everywhere downstream to properly interpret it.

jhe-:
Now that I think of it, using doubles is always within 1 nanosecond of a 128-bit integer divide, and that's due to the latter's truncation. It has 52 bits of precision so the value is basically dead on, for up to 9.2 billion seconds (2^63/1e9) of system uptime. That's longer than this country has existed.

codemercenary:
Yep, I did that math out as well.  Surprisingly often in theoretical physics you need to multiply one extremely large number by another extremely small number; a typical solution is to extract logarithms and do the math out that way; the approach we are using here is basically the same.

codemercenary:
It's OK if the quantity `1e9*mod_inv(g_Frequency)` overflows.  The real problem is that the modular inverse trick only works for exact division; if you want standard quotient/remainder division, you need a more complex multiplication.

jhe-:
Dude. It's got 99 problems, but overflowing ain't one.

I didn't say overflowing was an issue, I just meant that you cannot interpret (1e9*mod_inv(g_Frequency)) meaningfully without g_Frequency itself.

jhe-:
This is turning out to be the exact opposite of a clever trick ;)

jhe-:
Yeah, logarithms is definitely the right way to think of it.

codemercenary:
There is no better way to do this, they all wind up being the same thing, and 52 bits of precision is more than enough for our purposes.

PR #419, opened 2015-02-18
Manually create AutowiringEvents

gtremper:
AutowiringEvents is sometimes missing from the registry, causing problems for clients.


jnguyen75:
tested locally and it fixed my issue.  Thanks!

codemercenary:
I can't believe how many times we had to fix this problem, it seems like every time we thought it was good it just kept coming back.

Someday we'll deprecate Autowiring events completely, then we can obviate most of the issues pertaining to the type registry.

PR #435, opened 2015-02-25
Fix libstdc++

gtremper:
Various build problems fixed on libstdc++ and arm


codemercenary:
Tests pass on Windows, we're good to go.

PR #438, opened 2015-02-25
ObjectTraits is now CoreObjectDescriptor

codemercenary:
We are preparing to make `CoreObjectDescriptor` a public interface, thus changing the name of this type is necessary.


jnguyen75:
compiles and tests pass on Mac.  Will merge when Travis is green

codemercenary:
Travis is taking a really long time to get webhooks from github.  It might be awhile before this one completes.

codemercenary:
Override safeties for this one.  Travis seems to be down and too much depends on timely delivery of the autowiring-0.4.4 release.

PR #439, opened 2015-02-25
Add runtime variants to Snoop and Unsnoop

codemercenary:
This feature allows users to describe members to be added to a context in a runtime way.  Lists of `CoreObjectDescriptor` instances may therefore be used to cross module boundaries and enable objects defined and implemented in other modules to snoop remotely defined contexts.

- [x] Tests pass on Windows
- [x] Tests pass on Linux
- [x] Tests pass on Mac

### DIFF
--- a/autowiring/CoreContext.h
+++ b/autowiring/CoreContext.h
@@ -247,7 +247,7 @@ protected:
   t_rcvrSet m_delayedEventReceivers;
   
   // Context members from other contexts that have snooped this context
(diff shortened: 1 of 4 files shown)


codemercenary:
We have to verify everything builds by hand while Travis is down.

codemercenary:
All green on target platforms.  Merge when ready.

jnguyen75:
LGTM

PR #443, opened 2015-02-26
CoreObjectDescriptor should be copyable

codemercenary:


jnguyen75:
compiles and tests pass on Mac.  Waiting for Travis

PR #449, opened 2015-03-05
Eliminate libstdc++ support from CMakeLists

### LINKED ISSUE #448: Remove libstdc++ support
The time has come...

codemercenary:
Closes #448 

The time has finally come to eliminate support for libstdc++ once and for all.  We will start by short-circuiting all preprocessor blocks and macros that make references to libcxx.  Then comes the slow, arduous process of fixing all of the stupid backflips that we had to make to classes such as `ObjectPool` to make them compatible with libstdc++.

libstdc++ is dead.  Long live libc++

### DIFF
--- a/AutowiringConfig.h.in
+++ b/AutowiringConfig.h.in
@@ -10,11 +10,3 @@
 
 // Building for ARM?
 #cmakedefine01 autowiring_BUILD_ARM
-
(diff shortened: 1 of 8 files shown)


gittyupagain:
libstdc++, for us, is only dead on OS X. It is continues to grow on Linux.

PR #451, opened 2015-03-09
Fix AutowiringEvents

jnguyen75:
Fixing the AutowiringEvents by manually checking the CoreObjectDescriptor since it may have also been manually added in in the JunctionBoxManager constructor

### DIFF
--- a/autowiring/CoreObjectDescriptor.h
+++ b/autowiring/CoreObjectDescriptor.h
@@ -2,6 +2,7 @@
 #pragma once
 #include "AnySharedPointer.h"
 #include "AutoFilterDescriptor.h"
+#include "AutowiringEvents.h"
 #include "BoltBase.h"
 #include "ContextMember.h"
 #include "CoreRunnable.h"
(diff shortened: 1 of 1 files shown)


autowiring/CoreObjectDescriptor.h
```diff
@@ -36,7 +37,11 @@ struct CoreObjectDescriptor {
     pFilter(autowiring::fast_pointer_cast<ExceptionFilter>(value)),
     pBoltBase(autowiring::fast_pointer_cast<BoltBase>(value)),
     receivesEvents(
-      [this]{
+      [this, value]{
+        // Because we manually added AutowiringEvents to the JunctionBoxManager, check here also
+        if (autowiring::fast_pointer_cast<AutowiringEvents>(value))
```
codemercenary: This shouldn't be necessary, it means that `AutowiringEvents` isn't appearing in the event type registry.  @gtremper, any thoughts?

gtremper: We've had problems where `AutowiringEvents` isn't added to the registry. A complete fix would guarantee its added to the registry so we can eliminate this quick fix, as well as [this hack](https://github.com/leapmotion/autowiring/blob/7c9312061fbcae4b81b93f27f0af1659d32c4969/src/autowiring/JunctionBoxManager.cpp#L19).

codemercenary: We need to guarantee that this type winds up in the registry or we will have other issues.

jnguyen75: How long would this investigation take?  Should I close this PR for now?

codemercenary: Would take too long, I will merge this one in the short run.

PR #454, opened 2015-03-12
Adding a slots/signals concept

codemercenary:


jnguyen75:
this will be handy for things like callback/notifications

PR #458, opened 2015-03-23
Fix autoconfig

wal-rus:
Fixes several issues with autoconfig, adds three minor features & some additional unit tests.


codemercenary:
Add `#include <stdexcept>` to `AutoConfigParser.cpp`

wal-rus:
ok to build

autowiring/AutoConfig.h
```diff
@@ -40,9 +40,24 @@ class AutoConfig:
...
+  {
+    // Register with config registry
+    (void)RegConfig<T, TKey...>::r;
+
+    if (!IsConfigured()){
+      m_manager->Set(m_key, T(std::forward<t_Args>(args)...));
+    }
+  }
+
+#if _WIN32 //Silence warnings about multiple default constructors on windows
```
codemercenary: Preprocessor macros are forbidden.  Must find another way to solve this problem.

src/autowiring/AutoConfigManager.cpp
```diff
@@ -105,6 +105,17 @@ void AutoConfigManager::AddCallback(const std::string& key, t_callback&& fx) {
   m_callbacks[key].push_back(fx);
 }
 
+void AutoConfigManager::AddCallback(t_add_callback&& fx) {
+  // Grab lock until done setting value
+  std::lock_guard<std::mutex> lk(m_lock);
+
+  for (auto& key : m_orderedKeys) {
```
codemercenary: Autowiring codebase does permit the omission of braces.

src/autowiring/AutoConfigManager.cpp
```diff
@@ -118,13 +129,21 @@ void AutoConfigManager::SetRecursive(const std::string& key, AnySharedPointer va
...
+
   // Actually set the value in this manager
   SetInternal(key, value);
   
   // Mark key set from this manager
-  m_setHere.insert(key);
+  auto inserted = m_setHere.insert(key);
+  if (inserted.second) {
+    m_orderedKeys.push_back(key);
+    for (auto& callback : m_addCallbacks) {
```
codemercenary: Can we ensure that entries will not be removed from `m_addCallbacks`?

src/autowiring/demangle.cpp
```diff
@@ -7,28 +7,36 @@
...
+std::string autowiring::demangle(const std::type_info& ti, bool* success) {
   int status;
   std::unique_ptr<char, void(*)(void*)> res{
     abi::__cxa_demangle(ti.name(), nullptr, nullptr, &status),
     std::free
   };
+  if(success != nullptr)
+    *success = status == 0;
+  
   return std::string(status == 0 ? res.get() : ti.name());
```
codemercenary: Recommend we change the behavior of this routine to return an empty string when demangling is impossible, rather than extending the parameter space

src/autowiring/AutoConfigManager.cpp
wal-rus: yes

src/autowiring/demangle.cpp
wal-rus: This is an artifact of when I was attempting to manually interpret when demangling failed. will fix.

src/autowiring/AutoConfigManager.cpp
codemercenary: Excellent.

PR #464, opened 2015-03-25
Fix autorequired parentinject

wal-rus:


src/autowiring/test/ScopeTest.cpp
```diff
@@ -170,3 +170,109 @@ TEST_F(ScopeTest, AutowiringOrdering) {
   }
 }
 
+class InnerContext {};
+class MiddleContext {};
+TEST_F(ScopeTest, RequireVsWire) {
```
codemercenary: These tests are awesome.  +1 +1

PR #466, opened 2015-03-26
Add ContextEnumeratorT::unique and a default parameter

codemercenary:
Often, ContextEnumerator is used to obtain a single item.  Provide a helper routine which checks the single item assumption and also simplify the constructor syntax to make ContextEnumeratorT easier to use.


vdods:
Note: ContextEnumerator::iterator doesn't satisfy the requirements of the ForwardIterator concept (see http://en.cppreference.com/w/cpp/concept/ForwardIterator ), so it can't be used in std::next, for example.

PR #468, opened 2015-03-26
Make ContextEnumerator a ForwardIterator

codemercenary:
This should make `ContextEnumerator::iterator` a bit more flexible for algorithms which cannot operate on an `InputIterator` alone.


codemercenary:
Sure.

vdods:
Thanks!

PR #471, opened 2015-03-26
Fix initializer list ordering in SlotInformationStackLocation

gtremper:


codemercenary:
I really appreciate little things like this, I'm glad that you just fixed what you saw was broken whenever you had a spare moment.

PR #472, opened 2015-03-27
Add autoboost/stl interoperability

codemercenary:
`autoboost::future` is still being used on Android, make its `wait_for` method compatible with `std::chrono` datatypes.


codemercenary:
ok to build

PR #475, opened 2015-03-30
Feature signal upgrades

wal-rus:
-Makes signal_node_base more full featured
-Adds support for self-referencing lambdas which take a pointer to a node as their first argument
-Adds the ability for registration objects (aka nodes) to delete themselves without also holding a reference to the list they belong to
-Makes removing a node from a list it did not belong to throw an exception

### DIFF
--- a/autowiring/auto_signal.h
+++ b/autowiring/auto_signal.h
@@ -1,5 +1,7 @@
 // Copyright (C) 2012-2015 Leap Motion, Inc. All rights reserved.
 #pragma once
+#include "Decompose.h"
+
 #include <atomic>
 #include <functional>
 #include <memory>
@@ -51,12 +53,18 @@ namespace autowiring {
       /// If you call this function, you are assuming responsibility for the memory and 
       /// are expected to call delete on the node.
       /// </remarks>
-      void remove() {
+      /// <returns>
+      /// A pointer to itself for easier chaining of operations.
+      /// </returns>
+      signal_node_base* remove() {
         // Clear linkage
         if (this->pBlink)
           this->pBlink->pFlink = this->pFlink;
         if (this->pFlink)
           this->pFlink->pBlink = this->pBlink;
+
+        this->pBlink = this->pFlink = nullptr;
+        return this;
       }
     };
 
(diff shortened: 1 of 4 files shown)


wal-rus:
0 errors, first try!

autowiring/auto_signal.h
```diff
@@ -201,7 +224,9 @@ namespace autowiring {
     typedef internal::signal_node<Args...> registration_t;
     typedef std::function<void(Args...)> function_t;
 
-    registration_t* operator+=(function_t&& fn) { return *m_relay += std::move(fn); }
+    template<typename t_Fn>
+    registration_t* operator+=(t_Fn && fn) { return *m_relay += std::move(fn); }
```
codemercenary: Use `std::forward`, here, not `std::move`.  You can't know for sure that `t_Fn` is an rvalue reference because of rvalue reference collapse rules.

autowiring/auto_signal.h
```diff
@@ -27,16 +27,33 @@ namespace autowiring {
         pBlink(nullptr)
       {}
 
-      virtual ~signal_node_base(void) {}
+      /// <summary>
+      /// Removes this node from the list it's in, or throws if it's the head.
+      /// </summary>
+      virtual ~signal_node_base(void) {
+       // Clear linkage
```
codemercenary: Recommend you do not unlink from the destructor.  There are no guarantees afforded w.r.t. synchronization here, and you cannot add them after-the-fact if you decide you need them later.

autowiring/auto_signal.h
```diff
@@ -44,10 +61,24 @@ namespace autowiring {
       signal_node(std::function<void(Args...)>&& fn) :
         fn(std::move(fn))
       {}
-
+      
+      //Functions where the first argument is a signal_node<...> or base type are also ok.
+      signal_node(std::function<void(signal_node<Args...>*, Args...)>&& newFn) :
+      fn([this, newFn](Args... args){ newFn(this, args...); })
```
codemercenary: Nice rewrite.

In this case, recommend you consider adding another `std::function` as a member variable rather than capturing `newFn` in your closure.  If you do that, then you can use `std::move` to transfer `newFn` to your member variable, potentially making it a bit faster.  Also it's a little clearer.

src/autowiring/test/AutoSignalTest.cpp
```diff
@@ -122,3 +122,121 @@ TEST_F(AutoSignalTest, RaiseASignalWithinASlotTest) {
...
+  bool handler_called1 = false;
+  int magic_number = 123;
+
+  //The main test is just if this thing will compile
+  signal_t::registration_t* registration1 =
+  signal1 += [&](autowiring::internal::signal_node_base* reg, int magic) {
+    ASSERT_EQ(magic, magic_number);
+    ASSERT_EQ(registration1, reg);
+    handler_called1 = true;
+  };
```
codemercenary: Tab this part out so people don't miss the fact that lines 192 to 197 are all the same statement.

src/autowiring/test/AutoSignalTest.cpp
```diff
@@ -122,3 +122,121 @@ TEST_F(AutoSignalTest, RaiseASignalWithinASlotTest) {
...
+
+  signal1(magic_number);
```
codemercenary: This is awesome.  Asymmetric invocation with concept check--I'm surprised you got this implemented over a weekend.

codemercenary:
Top score

autowiring/auto_signal.h
wal-rus: If I did that, then I'd also have to add a Call Function method and support structure so that the correct one would be called for each node.

src/autowiring/test/AutoSignalTest.cpp
wal-rus: I did it on the car ride down to my friend's party...

codemercenary: Triple bonus points.

PR #479, opened 2015-03-31
Add Barrier functionality

codemercenary:
Add Barrier functionality

Users can now infer progress on a `DispatchQueue` by making use of the `DispatchQueue::Barrier`.  This allows users to replace this:

    std::mutex lock;
    std::condition_variable cv;
    bool cond = false;
    *q += [&] {
        std::lock_guard<std::mutex> lk(lock);
        cond = true;
        cv.notify_all();
    };
    std::unique_lock lk(lock);
    cv.wait_for(lk, std::chrono::seconds(5), [&] { return cond; });

With this:

	q.Barrier(std::chrono::seconds(5));

`DispatchQueue::Barrier` also has improved performance because it does not require the user construct an explicit condition variable and mutex every time barrier support is required.

### DIFF
--- a/autowiring/DispatchQueue.h
+++ b/autowiring/DispatchQueue.h
@@ -187,6 +187,17 @@ public:
   void AddExisting(DispatchThunkBase* pBase);
 
   /// <summary>
+  /// Blocks until all dispatchers on the DispatchQueue at the time of the call have been dispatched
+  /// </summary>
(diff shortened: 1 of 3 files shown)


jnguyen75:
would it be too much to ask for a ```void Barrier()``` function that doesn't take a time argument and just blocks indefinitely?

codemercenary:
I will add one.

PR #482, opened 2015-04-02
Add timeout-free Barrier variant

codemercenary:
This `Barrier` variant is identical to the timed version, except that it does not time out.


src/autowiring/DispatchQueue.cpp
```diff
@@ -193,6 +193,21 @@ bool DispatchQueue::Barrier(std::chrono::nanoseconds timeout) {
...
+  *this += [&] { complete = true; };
+
+  // Obtain the lock, wait until our variable is satisfied, which might be right away:
+  std::unique_lock<std::mutex> lk(m_dispatchLock);
+  m_queueUpdated.wait(lk, [&] { return m_aborted || complete; });
+  if (m_aborted)
+    // At this point, the dispatch queue MUST be completely run down.  We have no outstanding references
+    // to our stack-allocated "complete" variable.  Furthermore, after m_aborted is true, no further
+    // dispatchers are permitted to be run.
+    throw autowiring_error("Dispatch queue was aborted while a barrier was invoked");
```
jnguyen75: should this be ```dispatch_aborted_exception```?  If not, what's the difference?

PR #485, opened 2015-04-02
Add an interface support query ability to CoreContext

codemercenary:
This allows users to get all implementors of a particular interface in a context.

This method returns an empty vector if it is called before the context is initiated.


gtremper:
Fix documentation related to `vector` and I'll merge

PR #486, opened 2015-04-02
Fix DispatchQueue Exceptions

jnguyen75:
Fixes:
* ```dispatch_aborted_exception``` should inherit from ```autowiring_error```
* ```DispatchQueue::Barrier(*)``` functions should throw ```dispatch_aborted_exception```

### DIFF
--- a/src/autowiring/DispatchQueue.cpp
+++ b/src/autowiring/DispatchQueue.cpp
@@ -82,7 +82,7 @@ void DispatchQueue::Abort(void) {
 void DispatchQueue::WaitForEvent(void) {
   std::unique_lock<std::mutex> lk(m_dispatchLock);
   if (m_aborted)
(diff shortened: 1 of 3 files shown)


src/autowiring/DispatchQueue.cpp
```diff
@@ -80,12 +82,12 @@ void DispatchQueue::Abort(void) {
 void DispatchQueue::WaitForEvent(void) {
   std::unique_lock<std::mutex> lk(m_dispatchLock);
   if (m_aborted)
-    throw dispatch_aborted_exception();
+    throw dispatch_aborted_exception("Dispatch queue was aborted while waiting for an event");
```
codemercenary: if you can differentiate exceptions with different error strings, do so.  Makes it a lot easier to track exceptions to a line number.

jnguyen75: Added a slight tweak for exceptions prior vs during wait

PR #491, opened 2015-04-10
Always use visibility hidden

gtremper:
Requested by @gittyupagain. Fixes problems when linking multiple libraries that use autowiring


codemercenary:
Funny that we were still addressing linker issues a full year after splitting Autowiring into its own library.  I guess it's not really possible to ever really be free of packaging concerns.

PR #498, opened 2015-04-23
Provide more descriptive error text for junction box misses

codemercenary:
This error is expected to arise often enough that we should try to ensure it's informative when it does appear.  In this case, more information is better.


codemercenary:
Thanks, @gtremper

PR #501, opened 2015-05-07
Bugfix context enumerator t unique

vdods:
Wrote a unit test exposing a bug in ContextEnumeratorT::unique.  Fixed said bug.


codemercenary:
+1

PR #505, opened 2015-05-15
Fix bug in ContextCreator Clear

wal-rus:
Resolves an issue where teardown notifiers which free a shared pointer to their enclosing context can cause a reentrant lock

### DIFF
--- a/autowiring/ContextCreatorBase.h
+++ b/autowiring/ContextCreatorBase.h
@@ -24,14 +24,16 @@ protected:
   /// <param name="ctr">The collection of contexts to be cleared</param>
   /// <param name="iter">An iterator locker, which can obtain a shared pointer from an iterator</param>
   /// <remarks>
-  /// This method is synchronized on the contextLock and this will generally make it thread-safe
+  /// This method is synchronized on the contextLock and this will generally make it thread-safe.
+  /// It will clear the local context list immediately, but it is not garunteed that the context
+  /// list will be empty at the time of return.
   /// </remarks>
   template<class Ctr, class Fx>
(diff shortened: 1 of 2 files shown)


autowiring/ContextCreatorBase.h
```diff
@@ -28,31 +28,17 @@ class ContextCreatorBase {
...
-      std::lock_guard<std::mutex> lk(m_contextLock);
-      for(const auto& ctxt : ctr) {
-        auto locked = locker(ctxt);
-        if(locked)
-          locked->SignalShutdown();
-      }
-      ctr.clear();
-      return;
+    // Copy out under lock
+    Ctr ctrCopy = (std::lock_guard<std::mutex>(m_contextLock), std::move(ctr));
```
codemercenary: +1

codemercenary: Recommend you change `ctrCopy` to `ctrLocal`, and update your comment to reflect the fact that what's happening here is a move, not a copy.

src/autowiring/test/ContextCreatorTest.cpp
```diff
@@ -200,3 +200,35 @@ TEST_F(ContextCreatorTest, VoidKeyType) {
...
+
+struct mySigil {};
+
+class BackReferencingClass:
+  public CoreRunnable
+{
+public:
+  bool OnStart(void) override {
+    AutoCurrentContext ctxt;
+    m_parentContext = ctxt->GetParentContext();
```
codemercenary: Why the parent context and not the current context?  Can you add a comment here to describe your reasoning?

codemercenary:
Good find and good fix, @wal-rus.  Thanks for the assist.

PR #507, opened 2015-05-16
Fix zero-length query call for DispatchQueue::Barrier

### LINKED ISSUE #504: Calling DispatchQueue::Barrier() with 0 timeout returns false
When playing with ```DispatchQueueTest.Barrier``` line 77, the unit test would fail with ```std::chrono::nanoseconds``` was 1e4, but pass with 1e5

codemercenary:
When Barrier is called with a value of zero, the return value must be set according to the number of elements in the queue as of the time of the call.  Not only is this a good optimization, but it's also a way to allow consumers to easily poll the status of the queue.

Fixes #504


jnguyen75:
thanks!

PR #511, opened 2015-05-17
Refactor DispatchQueue to use an intrusive list instead of an std::list

codemercenary:
An intrusive linked list, in this case, is a much better choice.  Heap allocations for each entry in the collection are unavoidably necessary because entries have a variable size, which means a linked list is an improvement over `std::list`.

If necessary, a next step could be to use a journal-based arena allocator in order to keep links in the list sequential with respect to each other.

- [x] Merge #507

### DIFF
--- a/src/autowiring/test/DispatchQueueTest.cpp
+++ b/src/autowiring/test/DispatchQueueTest.cpp
@@ -96,6 +96,22 @@ TEST_F(DispatchQueueTest, Barrier) {
   ASSERT_TRUE(ct->Barrier(std::chrono::seconds(5))) << "Barrier did not return even though a dispatcher should have completed";
 }
 
+TEST_F(DispatchQueueTest, BarrierWithSingleDelayedDispatcher) {
+  AutoCurrentContext()->Initiate();
+  AutoRequired<CoreThread> ct;
+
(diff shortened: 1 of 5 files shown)


src/autowiring/test/DispatchQueueTest.cpp
```diff
@@ -118,16 +127,14 @@ TEST_F(DispatchQueueTest, BarrierWithAbort) {
...
   );
 
-  // Delay for long enough for the barrier to be reached:
-  std::this_thread::sleep_for(std::chrono::milliseconds(1));
-
   // Now abandon the queue, this should cause the async thread to quit:
   ct->Abort();
   ASSERT_EQ(std::future_status::ready, f.wait_for(std::chrono::seconds(5))) << "Barrier did not abort fast enough";
-  ASSERT_TRUE(*exception) << "Exception should have been thrown inside the Barrier call";
+  ASSERT_FALSE(f.get()) << "Exception should have been thrown inside the Barrier call";
```
gtremper: This assert fails on mac

codemercenary:
Added a test to make the failure @gtremper identified occur more reliably, then added another test and corrected said behavior.

PR #516, opened 2015-05-26
Crash in AutoSignal when Autowiring a member that uses virtual inheritance

wal-rus:
- [ ] Confirm that these tests pass on Windows


codemercenary:
Is this ready to merge @wal-rus ?

wal-rus:
Not quite.  Somehow the build is passing travis, but fails locally - it may be a windows only error

PR #526, opened 2015-06-03
Fix dispatch-abort lambda issue

codemercenary:
Fix yet another crazy reentrancy countercase.  Sequence of events is follows:

0. Create a dispatch queue `dq`
1. Create a type `Foo` that calls `dq.Abort` in its dtor
2. Create an instance `foo` of type `Foo`
3. Capture `foo` in a lambda, pend that lambda to `dq`

This causes a crash.  Invoking `d.operator+=` instead of `d.Abort` can cause the same issue under certain circumstances.  Fix by ensuring we don't hold locks while deleting dispatchers.


codemercenary:
Hangs as expected.

PR #546, opened 2015-06-09
Add a ThreadPool concept for CoreContext

codemercenary:
The thread pool concept allows users to attach a lambda to the context to be executed by that context's thread pool.  By default, contexts receive the thread pool of their parent, and the global context gets the default thread pool for the system.  Contexts assign their thread pool to the thread pool of their parent upon initiation (not construction).  Prior to initiation, contexts hold the null thread pool, which is an instance of NullPool that simply accumulates dispatchers until the parent context is initiated, shut down, or destroyed.

Setting a context's thread pool will not cause the context to release its reference to the NullPool.  Instead, the provided thread pool will be held until the context is started, and when this happens, all dispatchers will be transferred to the requested pool and then the requested pool will be initiated.

Jobs added to a context's thread pool will not be run until the context is started.

- [x] Builds and tests pass on Windows x86
- [x] Builds and tests pass on Windows x64
- [x] Builds and tests pass on Mac

### DIFF
--- a/autowiring/CoreContext.h
+++ b/autowiring/CoreContext.h
@@ -21,6 +21,7 @@
 #include "CoreObjectDescriptor.h"
 #include "result_or_default.h"
 #include "TeardownNotifier.h"
+#include "ThreadPool.h"
 #include "TypeRegistry.h"
 #include "TypeUnifier.h"
 
@@ -57,6 +58,8 @@ class JunctionBox;
 namespace autowiring {
   template<typename... Args>
   struct signal_relay_t;
+
+  class ThreadPool;
 }
 
 /// \file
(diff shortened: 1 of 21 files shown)


gtremper:
I have 4 failing tests on Mac:
* `ThreadPoolTest.SimpleSubmission`
* `ThreadPoolTest.PoolOverload`
* `ThreadPoolTest.StlPoolTest`
* `ThreadPoolTest.PendBeforeContextStart`

Which is every `ThreadPoolTest` except for `ManualThreadPoolBehavior`. They seem to be failing inconsistently as well. If I run only the `ThreadPoolTest.*` tests, they all seem to pass.

codemercenary:
Can you print the failure messages?

gtremper:
The appended lambdas seem to not run consistently. They tests fail because the `wait_for` call on the promise times out.  I'm going to lower the timeout time because 30 seconds makes testing this tedious

codemercenary:
Sure thing, thanks for the help.

codemercenary:
Lowering it to 5 should be fine.

codemercenary:
OK, I'm able to get the same failure.  I'm looking into this now.

PR #559, opened 2015-06-11
Correct potential memory smasher in GetAll

codemercenary:
`std::get_temporary_buffer` does not guarantee sufficient space for allocations in the same way that `new` does.  Use `new` here instead in order to ensure we do not smash memory during `GetAll`.


codemercenary:
I couldn't get `make_unique` working on all platforms--but if it does, then sure, why not?

jdonald:
Ran valgrind on MultiPostHocIntroductionTest, MultiDecorateTest, UnsatDecTest which call GetAll(). Doesn't report any problems for this branch.

codemercenary:
Thanks @jdonald !

PR #562, opened 2015-06-12
Use autowiring noop instread of empty lambda to expand parameter packs

gtremper:
I think this more clearly shows what's happening

### DIFF
--- a/autowiring/CoreContext.h
+++ b/autowiring/CoreContext.h
@@ -9,6 +9,7 @@
 #include "Bolt.h"
 #include "CoreRunnable.h"
 #include "ContextMember.h"
+#include "CallExtractor.h"
 #include "CreationRules.h"
 #include "CurrentContextPusher.h"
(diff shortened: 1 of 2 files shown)


codemercenary:
+1

autowiring/CoreContext.h
```diff
@@ -290,7 +291,7 @@ class CoreContext:
   // Enables a boltable class
   template<typename T, typename... Sigils>
   void EnableInternal(T*, Boltable<Sigils...>*) {
-    [](...){}((AutoRequireMicroBolt<T, Sigils>(),false)...);
+    autowiring::noop((AutoRequireMicroBolt<T, Sigils>(),false)...);
```
codemercenary: Do we still need the `,false` part of this?

gtremper: I tried removing the `,false` and it didn't compile. You can't pass `void` as an argument to a function.

PR #575, opened 2015-06-15
Add Autowiring diagnostic routines

codemercenary:
These routines are intended to be used to diagnose Autowiring issues from a debugger that is capable of evaluating function calls.  None of these routines make stateful changes to the system.  These methods assume that the debugger will run these functions from the thread context current in the debugger.


codemercenary:
@gtremper I might need your help with this one, I think the issue is down to the string parsing for type names and this will probably need debugging on the platform where the error is happening.

PR #581, opened 2015-06-17
Add debug function to print context tree

gtremper:


src/autowiring/AutowiringDebug.cpp
```diff
@@ -69,6 +69,45 @@ std::string autowiring::dbg::ContextName(void) {
...
+        os << "│   ";
+      else
+        os << "    ";
+
+    // Print context name
+    os << autowiring::demangle(ctxt->GetSigilType());
+    if (ctxt == curCtxt) {
+      os << "(Current Context)";
+    }
+    os << "\n";
```
codemercenary: Recommend you use `std::endl` here instead, it has a side-effect of performing `std::flush`.

src/autowiring/AutowiringDebug.cpp
```diff
@@ -69,6 +69,45 @@ std::string autowiring::dbg::ContextName(void) {
...
+    PrintContextTreeRecursive(os, ctxt->FirstChild());
+  }
+}
+
+void autowiring::dbg::PrintContextTree(std::ostream& os) {
+  PrintContextTreeRecursive(os, AutoGlobalContext());
+  os << std::flush;
```
codemercenary: The above recommendation would free you from having to perform an explicit `std::flush` here.

PR #582, opened 2015-06-18
Make CoreContext::Create<T> return the most-specific type possible.

vdods:
This makes CoreContext::Create<T> (and related functions) return std::shared_ptr\<CoreContextT\<T>> instead of just std::shared_ptr<CoreContext>.

- [ ] Wait of victor to give the go

### DIFF
--- a/autowiring/CoreContext.h
+++ b/autowiring/CoreContext.h
@@ -586,7 +586,7 @@ private:
   /// Creation helper routine
   /// </summary>
   template<class T>
-  static std::shared_ptr<CoreContext> UntypedCreate(
+  static std::shared_ptr<CoreContext> CreateUntyped(
     std::shared_ptr<CoreContext> pParent,
     t_childList::iterator backReference
   ) {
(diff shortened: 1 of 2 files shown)


vdods:
@codemercenary The placement of the `private` method in CoreContext is probably incorrect, but I wasn't sure what the correct place was.  Also, can you review the unit test?  Also, I don't currently know enough about AutoInjectable to write a meaningful unit test for the CoreContext::Create<T> method which accepts an AutoInjectable&&.

vdods:
Hold off on this for the moment.

codemercenary:
Good to go now?

vdods:
If you're ok with the placement of the private method UntypedCreate (adjacent to the public method Create), then it's ready.

autowiring/CoreContext.h
```diff
@@ -573,23 +573,35 @@ class CoreContext:
...
+    return std::make_shared<CoreContextT<T>>(pParent, backReference);
+  }
+
+private:
+  /// \internal
+  /// <summary>
+  /// Creation helper routine
+  /// </summary>
+  template<class T>
+  static std::shared_ptr<CoreContext> UntypedCreate(
```
codemercenary: Prefer the name `CreateUntyped` instead.  This causes all `Create` methods to be near each other in sorted statement completion lists.

vdods: Done.  Good point.

PR #588, opened 2015-06-18
Use iterators to check if last element

gtremper:
Shouldn't need to depend on the `operator==` of the `value_type` of the container.

### DIFF
--- a/src/autowiring/AutowiringDebug.cpp
+++ b/src/autowiring/AutowiringDebug.cpp
@@ -82,13 +82,13 @@ void PrintContextTreeRecursive(std::ostream& os, std::shared_ptr<CoreContext> ct
     }
 
     // Print indentation the shows 'ctxt's path in the tree
-    for (auto c : path)
-      if (c == path.back())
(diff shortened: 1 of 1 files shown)


src/autowiring/AutowiringDebug.cpp
```diff
@@ -82,13 +82,13 @@ void PrintContextTreeRecursive(std::ostream& os, std::shared_ptr<CoreContext> ct
     }
 
     // Print indentation the shows 'ctxt's path in the tree
-    for (auto c : path)
-      if (c == path.back())
-        if (c->NextSibling())
+    for (auto it = path.begin(); it != path.end(); ++it)
+      if (std::next(it) == path.end()) // Check if last element
```
codemercenary: Because we're iterating a concurrent collection, it is possible for `std::next(it) == std::next(it)` to evaluate `false`.  You might need to consider keeping the value of the next position around to ensure traversal happens as you expect or you might wind up printing a "├──" where you meant to print "└──".

PR #589, opened 2015-06-19
Prevent pointer types from being decorated onto an AutoPacket.

vdods:
Added a static_assert which prohibits types T satisfying std::is_pointer<T>::value from being decorated onto an AutoWiring packet, because the ownership/lifetime issue of said pointer is not a concern that we want to deal with.

### DIFF
--- a/autowiring/AutoPacket.h
+++ b/autowiring/AutoPacket.h
@@ -390,6 +390,7 @@ public:
   /// </remarks>
   template<class T>
   const T& Decorate(T t) {
+    static_assert(!std::is_pointer<T>::value, "Can't decorate using a pointer type.");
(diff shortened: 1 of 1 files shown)


vdods:
I verified (by hand in an ad-hoc manner) that the static_assert is indeed triggered in the correct case (e.g. by attempting to decorate int* onto a packet), while still the lib/tests build and tests pass.

codemercenary:
Does platform still compile?

vdods:
Good question.  I'll try it.

vdods:
Platform compiles with this build.  I also verified that it correctly triggers the static_assert if a pointer type is passed in within platform code (via temporarily adding an offending Decorate call).

PR #593, opened 2015-06-19
Added basic_timer.h and chrono_types.h

wal-rus:
Adds extremely useful convenience methods for using std::chrono


codemercenary:
I get bit by the linter every damn time.  Thanks for that, @gtremper.

PR #597, opened 2015-06-19
Initial add of Autowiring tutorials (AutoFilter)

vdods:
This adds 2 tutorials for AutoFilter (which are also compilable and executable C++) which produce pages in the doxygen-generated pages.

Currently there are some issues with the HTML generated for the tutorial pages, including a difference between public and non-public doxygen docs generation.  Running doxygen 1.8.8 by hand on Doxyfile (not publicDoxyfile.conf) produced the least-issue-having HTML docs for me.


examples/AutoFilterTutorial.0.0.cpp
```diff
@@ -0,0 +1,142 @@
+// AutoFilterTutorial.0.0.cpp created by Victor Dods on 2015/06/03
```
codemercenary: Git history already documents the creator of a file, there's no need to place a redundant contributor notice here.

examples/AutoFilterTutorial.0.0.cpp
```diff
@@ -0,0 +1,142 @@
...
+// Copyright (C) Leap Motion, Inc. All rights reserved.
```
codemercenary: Standard requires that this be the first line--that's why the build server is failing right now.  Also we have a standard form for this notice, it looks like this:

```C++
// Copyright (c) 2010 - 2013 Leap Motion. All rights reserved. Proprietary and confidential.
```

examples/AutoFilterTutorial.1.0.cpp
```diff
@@ -0,0 +1,232 @@
+// AutoFilterTutorial.1.0.cpp created by Victor Dods on 2015/06/10
```
codemercenary: Same story here

PR #607, opened 2015-06-23
Add a multi-unsatisfiable unit test

codemercenary:
This unit test was drafted awhile back to guard against regressions in `AutoPacket::Unsatisfiable`, might as well add it rather than let it rot.


codemercenary:
Looks like this doesn't bring in any changes at all--closing this PR as pointless.

PR #610, opened 2015-06-23
Don't install our own version of cmake

gtremper:


gtremper:
It appears Travis still has cmake version < 3.0

PR #612, opened 2015-06-23
arm: remove autoboost headers

Okyara:
- [ ] Verify this still builds on Android

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,15 +190,6 @@ if(NOT AUTOWIRING_IS_EMBEDDED)
     FILES_MATCHING PATTERN "*.h"
   )
   
-  # Install autoboost headers on ARM, which still requires them
-  if(autowiring_BUILD_ARM)
-    install(
-      DIRECTORY ${PROJECT_SOURCE_DIR}/contrib/autoboost/autoboost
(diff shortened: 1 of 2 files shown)


jdonald:
This builds for aarch64 gcc 4.9, but not the 32-bit arm gcc 4.8, as we require boost::future.

If you want to take a step toward our gcc 4.9 r10d NDK upgrade, you could make this code conditional on the gcc version.

jdonald:
Talked to Oksana about this in person and now understand the purpose behind it. These were her local changes to build for ARM Linux.

What we really need to do is a simpler change: replace the autowiring_build_ARM condition with something like autowiring_build_ANDROID. That is, continue using autoboost for Android, but not use it for ARM Linux or anything else. Oksana will fix this soon.

jdonald:
Oh, we also need the corresponding changes to <tt>#cmakedefine01 autowiring_BUILD_ANDROID</tt> AutowiringConfig.h.in, and set the new CMake variable around line 15 of CMakeLists.txt

So this merge likely broke Android 32-bit.

codemercenary:
This pull request was incorrectly targeted at master.  I have done the merge again and pushed it to develop instead in commit ddf684a90

PR #621, opened 2015-06-27
Auto out

vdods:
Added the newly designed auto_out feature, which allows for autofilters to have deferred output, while avoiding the trapdoor of manual packet decoration.  The presence of auto_out<T> in the parameter list of an autofilter allows the output to be treated as first class, and in particular, discoverable and visualizable by the autofilter graph output debug tool.


codemercenary:
Closed, changes herein replicated and superceded by #625, #626, #627, #628, and #629

PR #623, opened 2015-06-29
Relax constraint requiring that AutoFilters have unique output types

codemercenary:
Should be able to declare an `AutoFilter` method that has the same argument type twice as an output without problems, as this behavior is already supported for lambda functions and behaves the way one might expect.

### DIFF
--- a/autowiring/has_autofilter.h
+++ b/autowiring/has_autofilter.h
@@ -50,30 +50,6 @@ struct has_autofilter_arity<W, false>:
   static const int N = -1;
 };
 
-//=============================================
-// Test whether AutoFilter has unique arguments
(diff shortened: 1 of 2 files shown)


jnguyen75:
Neat trick to count the args in test:
```*factory += [&] (const Decoration<0>* args []) { while (*args++) nArgs++; };```

Builds and tests pass on mac.  Can merge when Travis passes

**Comment by codemercenary** [comment]
Yep, I like the dereference-increment-test idiom.  When I'm iterating a string, I use the same trick--finding the length of a string, for instance:

```C++
for(size_t i = 0; *p++; i++);
```

PR #629, opened 2015-07-01
Move specializations of auto_arg to their respective types

codemercenary:
These specializations require access to AutoPacket arguments.  Move them to the types they specialize, where possible, to break a dependency cycle.


jhe-:
Manually checked that this is literally cut-paste. :+1:

PR #632, opened 2015-07-07
android: restore autoboost headers

jdonald:
The changes in #612 are suspect, using a macro called autowiring_BUILD_ANDROID that is never defined, effectively disabling autoboost and the workaround NanosecondsForFutureWait()

This branch restores autoboost to 32-bit gcc 4.8 Android, while helping 64-bit gcc 4.9 move forward to use the \<future> header without autoboost.

Tested <tt>make install</tt> for both configurations to confirm that autoboost installs, or doesn't, as intended.


codemercenary:
Thanks for looking into this, we will need to cut 0.7.1 fairly soon.

PR #633, opened 2015-07-10
Add a reference overload for WriteAutoFilterGraph

codemercenary:
Also switch this method to use `AutowiredFast`.


jnguyen75:
Is this a replacement for ```AutoPacketGraph```?

PR #634, opened 2015-07-10
Switch to using a container for Travis builds

codemercenary:
Switch to container builds for speed and performance reasons as per [this](http://docs.travis-ci.com/user/migrating-from-legacy) discussion.  Full VM builds are slow and unresponsive, this system is far superior.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,22 @@
 language: cpp
 compiler: g++
 os: linux
+sudo: false
+env:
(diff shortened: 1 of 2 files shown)


codemercenary:
Update:  Container builds are actually slightly slower, but they seem to kick off faster than their VM counterparts, and Travis seems to want us to move over there.  In any event, build times are at least something that's within our power to address.

PR #637, opened 2015-07-10
Eliminate race condition in ContextMapTest.VerifyWithThreads

codemercenary:
This race condition occurs because we weren't waiting for the context to actually finish terminating before attempting to assess the state of the context map.  This meant that the secondary thread was sometimes still running even when we expected it to have already terminated.


wal-rus:
+1 for adding better tests too.

PR #639, opened 2015-07-10
Speed up a slow unit test

codemercenary:
Tests that have to sleep are bad and are generally starting to impact the total amount of time it takes to run all of the unit tests.  Replace them with monitor objects wherever possible.


hham:
It works fine.

PR #645, opened 2015-07-11
Eliminate unneeded overload

codemercenary:
This overload incorrectly provides read/write access to an underlying type on the `AutoPacket`, and thus should be removed immediately without a deprecation schedule.


codemercenary:
Oops.  This wasn't meant to be opened.

PR #646, opened 2015-07-11
Add representational equivalence check

codemercenary:
A planned optimization involves minimizing the use of `AnySharedPointer` in favor of `std::shared_ptr<void>`.  To make this change, it will be necessary to ensure that this type can be reinterpret casted to `std::shared_ptr<T>` for all `T`.

Add a unit test to ensure that this cast is possible and well-behaved on all target platforms.


wal-rus:
Yay tests!

PR #651, opened 2015-07-15
Relax behavior on global context checks

codemercenary:
Some consumers presently use the `AutowiringEnclosure` type at the same time as they keep a global context around.  While not the best practice from a test partitioning standpoint, we are nevertheless compelled to support this use case until downstream clients are refactored to stop making use of it.


jdonald:
Until someone uses this feature, it looks harmless, a no-op for the default no-args AutowiringEnclosure constructor.

PR #656, opened 2015-07-22
Add a unit test expressly for AutoFilter descriptors

codemercenary:
Some of the planned shared pointer slot refactoring is breaking `AutoFilterDescriptor` behaviors.  Because this type is fairly low-level, it makes sense to separate out tests dealing with this type into their own test case in order to speed debugging.


jdonald:
Old test looks unchanged, and the new ArityCheck test passes for me on Linux x64.

PR #657, opened 2015-07-22
Fix deprecated method calls in unit tests and internal calls

codemercenary:
`AutoPacket::Unsatisfiable` has been marked as deprecated since 0.7.1, failing to refactor Autowiring to use the new method name was an oversight.


jdonald:
Looks like these method bodies are identical in AutoPacket.h

PR #659, opened 2015-07-23
Eliminate copy ctor for AutoPacketFactory

codemercenary:
This type cannot be safely copied as it is concurrent anyway, might as well make this explicit.


hham:
Looks good.

PR #660, opened 2015-07-23
Implement Windows XP fallback mechanism for thread pools

codemercenary:
This is necessary to ensure that compatibility libraries that use Autowiring are still dynamically linkable on Windows XP without delay loading.  Hopefully this is the last LH+ API that needs to be refactored.


jdonald:
SystemThreadPoolWinLH.cpp looks equivalent to the old functionality and the tests pass so I assume this is safe for Win7+. Will merge then test off develop to find the next XP roadblock if there is one.

PR #661, opened 2015-07-23
Add `autowiring::is_complete`

codemercenary:
This will be used to determine at compile time whether a particular type is fully defined.  The utility of this compile time trait is pretty high; right now it's tough to know whether we can use `typeid` safely because types substituted in template methods and classes are often forward-declared.  At some point, however, they _do_ become completely defined, and we'd like to be able to instantiate types that rely upon total definition when that happens.

- [ ] Fix the Windows build


codemercenary:
I don't see any way to get this to compile in MSVC right now, this will have to be revisited in the future.

PR #662, opened 2015-07-23
XP: Fix GetProcAddress() usage of CreateThreadpoolWork()

jdonald:
It was properly delay-loading kernel32.dll to find these entrypoints, but then using the non-GetProcAddress() calls.


codemercenary:
Oops, thanks for fixing that for me.

PR #667, opened 2015-07-24
Teach AutoBench how to accept multiple benchmarks

codemercenary:
It's handy to be able to run multiple benchmarks at once without having to specify all of them and doesn't require that substantial of a refactor.


hham:
Looks good.

Issue #674, opened 2015-07-27
Fix ObjectPool performance


codemercenary:
`ObjectPool` has terrible performance because we used `std::shared_ptr` to track objects that are in the cache, when really we should be using `std::unique_ptr`.  This design choice was made in order to support libstdc, but because we no longer support it, we should probably fix `ObjectPool`'s current performance issue, especially if we are recommending that people use it in preference of `operator new`.

codemercenary:
Addressed by #678

PR #676, opened 2015-07-28
Add a benchmark test for the object pool

codemercenary:
`ObjectPool` will need to be a performant type.  Add a benchmark so we can track its improvement.  Current object pool times show `ObjectPool` as about 4x slower than `malloc` for flat datatypes, and 2x slower than `malloc` for types that allocate additional memory.  This is clearly an unacceptable situation.

Addresses #674


codemercenary:
Current benchmarks:

```
basic                                   68.542  ns  100.00%
complex                                 219.18  ns  319.77%
pool_basic                              378.65  ns  552.43%
pool_complex                            383.64  ns  559.71%
```

PR #677, opened 2015-07-28
Fix AUTO_ALIGNOF copy/paste typo

gittyupagain:


codemercenary:
I will merge this separately, but note that this change is also present in #666.

PR #678, opened 2015-07-28
Improve object pool performance

codemercenary:
Reduce the number of things that have to be copied to perform an allocation to just two--the pool version and the monitor reference.

### DIFF
--- a/autowiring/ObjectPool.h
+++ b/autowiring/ObjectPool.h
@@ -46,25 +46,18 @@ public:
     const std::function<void(T&)>& initial = &DefaultInitialize<T>,
     const std::function<void(T&)>& final = &DefaultFinalize<T>
   ) :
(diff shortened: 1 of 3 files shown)


codemercenary:
This marks a pretty significant performance improvement for `ObjectPool`.  #676 shows the previous benchmarks; current benchmarks are as follows:

```
objpool
basic                                   67.7915 ns  100.00%
complex                                 215.58  ns  318.00%
pool_basic                              213.83  ns  315.42%
pool_complex                            212.82  ns  313.94%
```

While the object pool still cannot compete with the native system allocator for uninitialized memory allocation, it is nevertheless significantly faster when allocationg data structures with nontrivial constructors, even when said constructors are kept very simple.  For more complex allocators, `ObjectPool` is expected to outperform the system heap significantly.

Issue #681, opened 2015-07-28
Eliminate ObjectPool constructing allocator


codemercenary:
`ObjectPool` has a constructor overload which allows callers to override default memory allocation.  This makes two separate allocations necessary and can be very inefficient.  Remove the allocating constructor completely, and then use `alignof` to satisfy alignment requirements.

PR #684, opened 2015-07-28
Fix compatibility, implement ObjPool placement constructor

codemercenary:
We would like to deprecate the allocating constructor variant of `ObjectPool`, but we cannot do this all at once.  Encourage users to use a placement constructor rather than an allocating constructor.

### DIFF
--- a/autowiring/ObjectPool.h
+++ b/autowiring/ObjectPool.h
@@ -21,7 +21,7 @@ void DefaultFinalize(T&){}
 
 namespace autowiring {
   struct placement_t {};
-  static const placement_t placement;
+  static const placement_t placement{};
 }
 
 /// <summary>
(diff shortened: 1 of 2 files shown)


autowiring/ObjectPool.h
```diff
@@ -58,11 +65,19 @@ class ObjectPool
...
     const std::function<void(T&)>& final = &DefaultFinalize<T>
   ), "Superceded by the placement construction version");
-  
-  ObjectPool(ObjectPool&& rhs)
-  {
-    *this = std::move(rhs);
-  }
+
+  /// <param name="limit">The maximum number of objects this pool will allow to be outstanding at any time.</param>
+  /// <param name="maxPooled">The maximum number of objects cached by the pool.</param>
```
gittyupagain: The comments above appear to be stale.

codemercenary:
This is needed in preparation for #681.  Users have to have something to migrate to before we remove the original feature.

PR #691, opened 2015-07-29
Eliminate Autowired::GetContext

codemercenary:
Eliminate this low-level internal routine, as its use implies an antipattern.  Users instead should track their contexts themselves, and only use `Autowired` to access the type that's actually been wired in.  This change will allow us greater flexibility in deciding how to use `m_context` and when to reset it .


wal-rus:
Again, Verified this doesn't break anything obvious on windows.

codemercenary:
Mind merging?

PR #696, opened 2015-07-30
Support no-return parallel entries

### LINKED ISSUE #687: Enable parallel::begin<void>
Let the following code work:

```C++
for (int i : {0,4,2,5,1,3}) {
  int sleepTime = dist(mt);
  p += [i, sleepTime]() {};
}

std::vector<int> result;
for (auto it = p.begin<void(); it != p.end<void>(); ++it);
```

A specialization for `parallel::begin<void>` is needed to get this code functional.

codemercenary:
This allows lambdas that have no return type to be added to the parallel collection, and have their completion status be tracked the same way.

Fixes #687

### DIFF
--- a/autowiring/Parallel.h
+++ b/autowiring/Parallel.h
@@ -147,18 +147,14 @@ public:
   template<typename T>
   void Pop(void) {
     std::unique_lock<std::mutex> lk(m_queueMutex);
+    if (!m_outstandingCount)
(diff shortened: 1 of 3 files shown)


jdonald:
ParallelTest.VoidReturn passes, while the non-void tests e.g. ParallelTest.All still pass.

PR #698, opened 2015-07-30
Throw exceptions if the context isn't running during a call to begin()

### LINKED ISSUE #688: parallel::begin exception behavior
If a user attempts to call `parallel::begin` and the current context is not running, throw an exception.

codemercenary:
It's easy to make this mistake--because the main thread is often the one making the call to begin, do not allow users to block on a call to `autowiring::parallel::begin` until after the context is started.

Fixes #688


jdonald:
makes sense.

PR #707, opened 2015-08-03
Add arm build to travis

codemercenary:
We keep seeing regressions on ARM.  It's necessary we start building there or we're going to continue to see last-minute issues on that platform when we try to cut a release.  Need the following before this build will pass:

- [x] travis-ci/apt-package-whitelist#802
- [ ] travis-ci/apt-package-whitelist#850
- [ ] travis-ci/apt-package-whitelist#882
- [x] travis-ci/apt-package-whitelist#851
- [ ] 3x (arm, linux, mac) multi-platform build configured

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,26 +3,44 @@ sudo: false
 
 matrix:
   include:
+  # - os: linux
+    # env:
+    # - _CC=gcc-4.8
+    # - _CXX=g++-4.8
+    # - CMAKE_URL=http://www.cmake.org/files/v3.1/cmake-3.1.0-Linux-i386.tar.gz
+    # - CMAKE_DIRNAME=cmake-3.1.0-Linux-i386
+    # - ADDITIONAL_CMAKE_FLAGS=
+    # addons:
+      # apt:
+        # sources:
+          # - ubuntu-toolchain-r-test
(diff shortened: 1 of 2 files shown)


jdonald:
Looks like this image needs to run <tt>sudo apt-get -y install g++-arm-linux-gnueabihf</tt>

jdonald:
The above links seem to be about adding gcc-4.9-base in Travis whitelists? It seems like a gcc 4.9 upgrade would be independent (or rather complicate things) for supporting ARM cross-compiling.

It should be fine to use version 4.6 of arm-linux-gnueabihf-g++.

When it later comes to Android, we will need gcc 4.8 or newer as the STL support in NDK fall further behind, but ARM Linux should be fine on the same gcc we use for x86 Linux.

codemercenary:
Unfortunately we can't use `sudo` because we are using containerized builds on Travis.  I'd much rather get them to add this package to the whitelist than switch back to the legacy build system.

jdonald:
Ahh I see travis-ci/apt-package-whitelist#846 is for cpp-4.9-arm-linux-gnueabi

codemercenary:
Closing this as this is unlikely to ever happen in a billion goddamn years.

PR #711, opened 2015-08-06
Handle uninitialized block cases

codemercenary:
Sometimes an ID will be demangled that was never instantiated in the system.  Handle these cases by presenting an appropriate "unknown" indicator.


codemercenary:
Closing this as I think I have a better fix.

PR #712, opened 2015-08-06
Relax constraints on incomplete AutoFilter types

codemercenary:
Certain AutoFilter networks may consist entirely of incomplete types.  In this case, we do not have full type information, and should fall back to a synthetic type instead, and inform the user in the graph printout of the presence of an incomplete class.


hham:
Tests passed successfully!

codemercenary:
Indeed, feel free to merge when ready.

PR #715, opened 2015-08-10
Add an "observable" concept and corresponding unit test

codemercenary:
`observable` is a very useful pattern, and now that autowiring has signals, should be easy to add.  Implement an observable concept built as a wrapper around a value type.


hham:
This is simple and nice to use. Tests passed successfully.

PR #716, opened 2015-08-11
Restore run-compatibility with libc 2.15 / libstdc++ 3.4.15 systems

jdonald:
One of our users noticed that the Leap SDK Sample.cpp no longer links on Fedora 16, due to a dependency on std::thread::hardware_concurrency(). This would also affect Ubuntu 12.04 LTS.

This could be replaced with get_nprocs() on Linux, but for now it can be removed.

### DIFF
--- a/src/autowiring/SystemThreadPoolStl.cpp
+++ b/src/autowiring/SystemThreadPoolStl.cpp
@@ -44,7 +44,8 @@ void SystemThreadPoolStl::OnStartUnsafe(void) {
     // Do nothing if the pool size was already set by someone else
     return;
 
-  // TODO:  Set this number according to std::thread::hardware_concurrency.  This can't
+  // TODO:  Set this number according to std::thread::hardware_concurrency (or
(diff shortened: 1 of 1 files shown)


codemercenary:
Will keep this in mind.

codemercenary:
Note that this issue is unlikely to affect Ubuntu 12.04, given that this is what Travis is running and we test every single pull request on that platform.

jdonald:
Travis runs but also *builds* on Ubuntu 12.04, so it'll create entrypoints that are libc 2.15-compatible.

However, if we build our autowiring libraries on linbuild2 (Ubuntu 14.04), they can end up with run dependencies on glibc 2.19.

Another workaround (other than being careful on which functions we call as done in this PR)  is to make sure any autowiring libraries on linbuild2 are not built on linbuild2. Rather, build them on an older libc system such as Travis or ocuserv2 then upload to linbuild2. That's the only solution we currently have for libusb.

codemercenary:
Ah-hah.  That's a good distinction, I'll keep that in mind.

PR #717, opened 2015-08-11
Fix signals cyclic unlink issue

codemercenary:
Add unit test to show regression in cyclic unlink countercase.  This occurs when a signal handler is registered on an Autowired field which is unlinked by CoreContext.


wal-rus:
Verified that this resolves my issue.

PR #732, opened 2015-08-14
Makes events get unregistered before notifying teardown listeners.

codemercenary:
Teardown listeners should happen after everything else in the context is over.


codemercenary:
Closing this one, this causes test regresssions.

PR #734, opened 2015-08-22
Add a test of multi-in with single-in decoration

hham:
This PR add a test of AutoFilter with multi-in and single-in to show that a filter with multi-in and single-in is not allowed for now.


codemercenary:
Thank you, @hham !  This is awesome.

hham:
Sure.

jdonald:
So this is still broken on clang, but somehow works with gcc?

If I instantiate <tt>\[\&f_called](float\& out) { out = ++f_called; }</tt> before the multi-decoration, somehow the test passes.

codemercenary:
That is very strange indeed, a good finding too--should help me track this one down once I get some time to work on it.

codemercenary:
@jdonald would you mind merging this one?

hham:
Works fine on Unit tests on my Mac.

jdonald:
I double-checked that the test as originally written by Hyoung (floating-point variable and all), now passes on clang.

PR #739, opened 2015-09-12
Move spin_lock type from Platform to autowiring

codemercenary:
Autowiring is increasingly a concurrency library, this type has a better home here and could actually be used by some concepts that presently use the much more heavyweight `std::mutex`.


hham:
Works fine.

codemercenary:
OK, should be good to go now.

PR #740, opened 2015-09-17
Fix static assert in CreationRules

codemercenary:
Factory `New` methods presently are not being correctly evaluated for the inheritance rules that are required for factory `New` when the method has one or more arguments.  Alter the static assert to be applied in all cases where static new is used.


codemercenary:
That was not intentional, develop has been forced back and this PR resubmitted as #741

PR #741, opened 2015-09-17
Fix static assert in CreationRules

codemercenary:
Factory `New` methods presently are not being correctly evaluated for the inheritance rules that are required for factory `New` when the method has one or more arguments.  Alter the static assert to be applied in all cases where static new is used.


autowiring/CreationRules.h
```diff
@@ -85,7 +85,7 @@ struct crh<construction_strategy::factory_new, T, Args...>
   typedef T TActual;
   
   static_assert(
-    std::is_base_of<CoreObject, T>::value || !has_static_new<T>::value,
+    std::is_base_of<CoreObject, T>::value,
     "If type T provides a static New method, then the constructed type MUST directly inherit Object"
```
jdonald: This string "If type T provides a static New method..." matched the old static_assert, but if you make the above change, doesn't it become merely "The constructed type MUST directly inherit Object" ?

codemercenary: No, because this whole template type--`crh`--is only instantiated in cases where `T` is being constructed via a factory new call.

jdonald: I see now.

PR #742, opened 2015-09-17
Pull ThreadSleep into CoreRunnable

codemercenary:
Update `ThreadSleep` to use the condition variable in `CoreRunnable`, rather than the state block in `BasicThread`.  Also update the documentation on this method to reflect the method's actual behavior.


jdonald:
Unit tests pass (i.e. the two tests that use ThreadSleep(), WaitRunsDownAllPackets and EmptyPoolIssuance) still pass when I run them 1000x.

Merging.

PR #744, opened 2015-09-17
AutowiringEnclosure should print off hung runnables

codemercenary:
It's a bit easier to debug when things are hung when the enclosure actually prints off what's hung.  Make this change.


jdonald:
Awesome

PR #745, opened 2015-09-23
Fix XCode 7.0 Warnings as Errors

jnguyen75:
Accidentally upgraded my xcode so I may as well fix these warnings as errors


codemercenary:
IT BARFD

codemercenary:
Looks like Travis is in a bad state, #746 should fix the problem.

jdonald:
:+1: @jnguyen75 

There won't be an Xcode 6.5, so we will all be practically forced to upgrade soon.

PR #746, opened 2015-09-23
Fix Travis script

codemercenary:
Looks like CMake changed the location of their binary package download URLs to omit the www.  They did put in an HTTP 301 redirect (moved permanently), but curl will by default not follow such redirects, and so this is failing all of our builds.

Fix the URL to point to the new path, and also add a `-L` switch to curl so it will follow any redirects that happen in the future.


codemercenary:
Got to merge this one myself in order to fix the build server.

PR #747, opened 2015-09-23
Upkeep refactor of ContextMap

codemercenary:
Minor refactor to clean up a few idioms, nothing functionality changing here.


jdonald:
This is clearer, and I've confirmed ContextMapTest still passes consistently (valgrind-clean and all).

PR #749, opened 2015-09-24
Add and implement "once" concept

codemercenary:
There are lots of one-way transitions in the code that need to trigger other behavior, a "once" flag is a well overdue support concept to implement such transitions.

### DIFF
--- a/src/autowiring/test/OnceTest.cpp
+++ b/src/autowiring/test/OnceTest.cpp
@@ -49,7 +49,7 @@ TEST(OnceTest, MultiLambdaPending) {
   autowiring::once o;
   std::atomic<int> x{ 0 };
   std::atomic<int> y{ 0 };
-  bool go = false;
+  bool go = true;
 
   std::thread v([&] {
     while(go) {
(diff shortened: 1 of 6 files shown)


src/autowiring/test/OnceTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+  ASSERT_TRUE(v.unique()) << "Shared pointer leaked by once flag after execution";
+  o += [v] {};
+  ASSERT_TRUE(v.unique()) << "Shared pointer leaked by once flag after post-set attachment";
+}
+
+TEST(OnceTest, MultiLambdaPending) {
+  autowiring::once o;
+  std::atomic<int> x{ 0 };
+  std::atomic<int> y{ 0 };
+  bool go = false;
```
gittyupagain: Since the variable `go` is never set to `true`, `x` and `y` will always be `0` and equal in the final assert.

codemercenary: Crap

PR #751, opened 2015-09-26
Add dereference operator to observable

codemercenary:
Simple low-risk extension that should make this type more useful.


hham:
It works fine.

PR #753, opened 2015-09-26
Switch to a spin lock in auto_signal

codemercenary:
The signal type is an extremely low contention type that could benefit from not having to carry around an `std::mutex` everywhere.  Switch it out in favor of the newly introduced `autowiring::spin_lock`.


hham:
Tests were successful. Looks good. Now we've go faster signal.

PR #754, opened 2015-09-29
CoreContext should not define a copy ctor

codemercenary:
It would be an error for a CoreContext to be copied, and this should not be used anywhere by anyone.  Explicitly disallow it.


jdonald:
+1

PR #764, opened 2015-10-02
Add *_i32 and *i64 for fixed sized chrono types

wal-rus:
Useful if you only want an int, since the default is a long long.


hham:
Looks good.

PR #765, opened 2015-10-03
Fix valgrind AutowiringTest hang

jdonald:
Similar to #700 

valgrind's thread scheduling allows one thread to livelock forever unless it yields at some point.

### DIFF
--- a/src/autowiring/test/OnceTest.cpp
+++ b/src/autowiring/test/OnceTest.cpp
@@ -55,6 +55,8 @@ TEST(OnceTest, MultiLambdaPending) {
     while(go) {
       o += [&] { x++; };
       y++;
+      if (y % 10000 == 0)
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
     }
   });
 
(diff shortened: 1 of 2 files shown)


src/autowiring/test/OnceTest.cpp
```diff
@@ -55,6 +55,8 @@ TEST(OnceTest, MultiLambdaPending) {
     while(go) {
       o += [&] { x++; };
       y++;
+      if (y % 10000 == 0)
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
codemercenary: If you need to yield, recommend you do [`std::this_thread::yield()`](http://en.cppreference.com/w/cpp/thread/yield)

jdonald: When I change this to yield() it still hangs, likewise for a similar workaround in AutoSignalTest.. Apparently valgrind can choose to ignore yields.

Just now I learned that valgrind has an option `--fair-sched=yes` which better mimics reality and can get these tests to complete without sleep_for(). Given the nature of some these and some other autowiring unit tests, perhaps this should be mandatory for running AutowiringTest in valgrind.

codemercenary: That sounds like a good idea to me.  It's unlikely that this sleep will affect that much but I try to avoid using them where possible as a matter of policy.

What would you like to do

jdonald:
Closing this. Will file a separate PR to document `valgrind --fair-sched=yes bin/AutowiringTest` and remove calls to sleep_for()/WaitForEvent() that will no longer be necessary.

PR #767, opened 2015-10-05
Fix broken Unlink function

codemercenary:
Unlink wasn't actually removing the specified entry from the list.  What a dopy oversight.  Add a test so we don't make this error again.


jdonald:
Verified that this fixes the disconnect/reconnect crashes on OS X! Windows is still building but I'm optimistic that it'll work there too.

PR #768, opened 2015-10-05
Remove unnecessary sleeps and document valgrind --fair-sched=yes

jdonald:
Followup from #765 

By requiring `--fair-sched=yes`, we can use valgrind to catch invalid memory reads/writes, but need not continually put in sleep hacks as new pathological tests are added.


README.md
```diff
@@ -25,7 +25,7 @@ This will configure the project to build fat binaries by default.  If you wish t
 
     cmake . -DCMAKE_OSX_ARCHITECTURES:STRING=x86_64
 
-### Linix
+### Linux
```
codemercenary: Lol oops

PR #769, opened 2015-10-05
Add a test to assure post-hoc type advertisement behaviors are correct

codemercenary:
Intend to use this behavior in platform, need to make sure it works and stays working.

### DIFF
--- a/src/autowiring/test/AutowiringTest.cpp
+++ b/src/autowiring/test/AutowiringTest.cpp
@@ -99,6 +99,8 @@ class DynamicCastDerivedType : public PublicBase, private PrivateBase {
 public:
   DynamicCastDerivedType(void) {}
   ~DynamicCastDerivedType(void) {}
+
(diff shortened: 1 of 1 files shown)


src/autowiring/test/AutowiringTest.cpp
```diff
@@ -99,6 +99,8 @@ class DynamicCastDerivedType : public PublicBase, private PrivateBase {
 public:
   DynamicCastDerivedType(void) {}
   ~DynamicCastDerivedType(void) {}
+
+  FRIEND_TEST(AutowiringTest, PostHocTypeAdvertisement);
```
jdonald: Oh, that's a handy macro.

codemercenary: Yes but please, please resist the temptation to use this anywhere except in test code.  I hate it when unit tests infiltrate an otherwise pristine codebase.

PR #771, opened 2015-10-06
Add test to verify correct behavior on constructive injection

codemercenary:
Suspect that there is some misbehavior, here, when the same type is injected from multiple threads.  Add a test to uncover the issue, and then fix the problem.


hham:
Yes, I could verify that autowiring_error was thrown.

codemercenary:
Thank you @hham

hham:
No problem.

PR #780, opened 2015-10-15
Fix and test NotifyWhenAutowired call order

### LINKED ISSUE #738: NotifyWhenAutowired appears to be satisfied out of order
Example use case is with sliders, which use `NotifyWhenAutowired` on Figure. When I start LeapSvc with a device plugged in, a couple frames will tend to get processed before Figure comes online.

However, once Figure is autowired, these are often triggered in a different order than the `m_Params.NotifyWhenAutowired` line was called, which knocks sliders from their intended ordering.

This doesn't happen when I plug in a device after a few seconds, since at that point the calls to add sliders are processed immediately.

codemercenary:
`NotifyWhenAutowired` is meant to carry an outer total order guarantee which states that notifications pertaining to the injection of a particular type into a context must be processed strictly in the order they are received.

- [ ] Fixes #738

### DIFF
--- a/autowiring/AutowirableSlot.h
+++ b/autowiring/AutowirableSlot.h
@@ -68,9 +68,9 @@ protected:
   /// </remarks>
   std::weak_ptr<CoreContext> m_context;
 
-  /// The unique ptr to the registration of the singal handler which is used to finish
-  /// the autowiring. The registration can be used to cancel the signal handler.
-  std::unique_ptr<autowiring::registration_t> m_deferred_registration;
+  /// The registration of the singal handler which is used to finish the autowiring.
+  /// It can be used to cancel the signal handler.
+  autowiring::registration_t m_deferred_registration;
 
   /// Cancel the signal handler regsitered in m_deferred_regsitration. This is used when this
   /// is already autowired or to cancel autowiring.
--- a/autowiring/AutowirableSlot.h
+++ b/autowiring/AutowirableSlot.h
@@ -70,11 +70,11 @@ protected:
 
   /// The unique ptr to the registration of the singal handler which is used to finish
   /// the autowiring. The registration can be used to cancel the signal handler.
-  std::unique_ptr<autowiring::registration_t> m_deferred_autowire;
+  std::unique_ptr<autowiring::registration_t> m_deferred_registration;
 
-  /// Cancel the signal handler regsitered in m_deferred_autowire. This is used when this
+  /// Cancel the signal handler regsitered in m_deferred_regsitration. This is used when this
   /// is already autowired or to cancel autowiring.
-  void UnRegisterDeferredAutowire(void);
+  void UnregisterDeferredAutowire(void);
 
   /// <summary>
   /// Causes this deferrable to unregister itself with the enclosing context
--- a/autowiring/AutowirableSlot.h
+++ b/autowiring/AutowirableSlot.h
@@ -68,6 +68,9 @@ protected:
   /// </remarks>
   std::weak_ptr<CoreContext> m_context;
 
+  /// This is used to cancel autowiring
+  std::unique_ptr<autowiring::registration_t> m_deferred_autowire;
+
   /// <summary>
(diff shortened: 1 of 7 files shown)


src/autowiring/test/PostConstructTest.cpp
```diff
@@ -385,3 +385,22 @@ TEST_F(PostConstructTest, PostConstructCanSafelyThrow) {
   Autowired<PostConstructThrowsException> pcte;
   ASSERT_FALSE(pcte.IsAutowired()) << "A context member which threw an exception post-construction was incorrectly introduced into a context";
 }
+
+namespace {
+  class EmptyType : public CoreObject {};
+}
+
+TEST_F(PostConstructTest, StrictNotificationArrangement) {
+  size_t call[3] = {0, 0, 0};
```
codemercenary: Recommend you initialize these to something other than `0` so we can tell if the lambdas weren't even called.

veronicaz41: :+1:

autowiring/AutowirableSlot.h
```diff
@@ -68,8 +68,13 @@ class DeferrableAutowiring
   /// </remarks>
   std::weak_ptr<CoreContext> m_context;
 
-  // The next entry to be autowired in this sequence
-  DeferrableAutowiring* m_pFlink = nullptr;
+  /// The unique ptr to the registration of the singal handler which is used to finish
+  /// the autowiring. The registration can be used to cancel the signal handler.
+  std::unique_ptr<autowiring::registration_t> m_deferred_autowire;
```
codemercenary: Recommend you change this to `m_deferred_handler` or `m_deferred_registration`

autowiring/AutowirableSlot.h
```diff
@@ -68,8 +68,13 @@ class DeferrableAutowiring
...
+
+  /// Cancel the signal handler regsitered in m_deferred_autowire. This is used when this
+  /// is already autowired or to cancel autowiring.
+  void UnRegisterDeferredAutowire(void);
```
codemercenary: `Unregister`

autowiring/AutowirableSlot.h
```diff
@@ -119,20 +118,14 @@ class DeferrableAutowiring
   virtual DeferrableUnsynchronizedStrategy* GetStrategy(void) { return nullptr; }
```
codemercenary: Is this still needed?

autowiring/AutowirableSlot.h
codemercenary: Why did you have to make this a pointer type?  Is there a limitation with `autowiring::registration_t` that prevents you from making it a member?

autowiring/AutowirableSlot.h
veronicaz41: Not needed.

autowiring/AutowirableSlot.h
veronicaz41: Actually there's no limitation. Changed it to `autowiring::registration_t`.

codemercenary:
Looks like we got this one.  Merging.

autowiring/Autowired.h
```diff
@@ -112,13 +110,10 @@ class Autowired:
...
 
 private:
-  // The first deferred child known to need registration.  This is distinct from the m_pFlink entry in
-  // the base class, which refers to the _next sibling_; by contrast, this type refers to the _first child_
-  // which will be the first member registered via NotifyWhenAutowired.
-  std::atomic<AutowirableSlot<T>*> m_pFirstChild{nullptr};
-
   std::vector<autowiring::registration_t> m_events;
 
+  std::vector<autowiring::registration_t> m_autowired_notifications;
```
codemercenary: I don't like this `std::vector` instance, though, and would normally try to find a way to remove it.  On the other hand, there is another identical vector right above it, so this is probably not too bad.

PR #781, opened 2015-10-15
Add r-value support to AutoFilter

codemercenary:
Add support for R-value arguments to `AutoFilter`.  This argument type should make it possible to modify a decoration in-place, and should enforce necessary thread safety constraints to ensure that concurrent use and modification does not occur.

- [x] Add r-value support
- [x] Add test and fix recipient removal
- [x] Add test and fix altitude on packet
- [x] Make `AddRecipient` on `AutoPacket` idempotent, consistent with `AddSubscriber` on `AutoPacketFactory`
- [x] Add `AutoPacketGraph` and `AutowiringDebug` support for r-value

ToDo:
Refactor so we do not re-generate `m_decoration_map` every time we call `NewPacket()` on `AutoPacketFactory`. It is inefficient because we will detect cycle repeatedly.
Will make a separate pull request.


codemercenary:
I'd like to get this in for 0.7.6 if possible.  If not, we'll postpone.

codemercenary:
Bumping this off the milestone, we'll try for the next version.

autowiring/AutoPacket.h
```diff
@@ -573,13 +595,15 @@ class AutoPacket:
   /// Adds a recipient for data associated only with this issuance of the packet.
   /// </summary>
   /// <remarks>
-  /// This method is not idempotent.  The returned Recipient structure may be used to remove
-  /// the recipient safely at any point.  The caller MUST NOT attempt
```
codemercenary: Oh boy.  "MUST NOT attempt..." and nothing

autowiring/AutowiringDebug.h
```diff
@@ -10,62 +10,64 @@ class AutoPacket;
 class CoreContext;
 
 namespace autowiring {
-namespace dbg {
+  namespace dbg {
```
codemercenary: I like formatting fixups, but leave this for another PR.  It makes it harder to review important changes.

autowiring/DecorationDisposition.h
```diff
@@ -1,8 +1,10 @@
 // Copyright (C) 2012-2015 Leap Motion, Inc. All rights reserved.
 #pragma once
 #include "AutowiringConfig.h"
+#include "altitude.h"
 #include "AnySharedPointer.h"
 #include <cassert>
+#include <set>
```
codemercenary: Would `unordered_set` also be suitable?

autowiring/DecorationDisposition.h
```diff
@@ -103,24 +92,36 @@ struct DecorationDisposition
...
+    // The pointer to the satisfaction counter, it should never be nullptr
+    SatCounter* const satCounter;
+
+    bool operator<(const Subscriber& rhs) const {
+      return std::tie(altitude, satCounter) > std::tie(rhs.altitude, rhs.satCounter);
+    }
   };
 
-  std::vector<Subscriber> m_subscribers;
+  std::set<Subscriber> m_subscribers;
```
codemercenary: I'm reluctant to make this kind of change.  I understand that it has the advantage of making it easy to prevent reduplicated introductions, but that kind of guarantee should be made higher up, not here.

src/autowiring/AutoPacket.cpp
```diff
@@ -96,44 +100,115 @@ void AutoPacket::AddSatCounterUnsafe(SatCounter& satCounter) {
...
-            if (pOther->id == pCur->id && !pOther->is_multi) {
-              std::stringstream ss;
-              ss << "Added identical data broadcasts of type " << autowiring::demangle(pCur->id) << " with existing subscriber.";
-              throw std::runtime_error(ss.str());
+
+    if (pCur->is_rvalue) {
+      if (entry.m_pModifier) {
+        std::stringstream ss;
+        ss << "Added identical rvalue decorations for type " << autowiring::demangle(pCur->id);
+        throw std::runtime_error(ss.str());
```
codemercenary: I'm aware that `runtime_error` is thrown liberally elsewhere, but we should centralize our code to throw `autowiring_error` instead.

src/autowiring/AutoPacket.cpp
```diff
@@ -96,44 +100,115 @@ void AutoPacket::AddSatCounterUnsafe(SatCounter& satCounter) {
...
-          for (auto pOther = subscriber.satCounter->GetAutoFilterArguments(); *pOther; pOther++) {
-            if (pOther->id == pCur->id && !pOther->is_multi) {
-              std::stringstream ss;
-              ss << "Added identical data broadcasts of type " << autowiring::demangle(pCur->id) << " with existing subscriber.";
-              throw std::runtime_error(ss.str());
+
+    if (pCur->is_rvalue) {
+      if (entry.m_pModifier) {
+        std::stringstream ss;
+        ss << "Added identical rvalue decorations for type " << autowiring::demangle(pCur->id);
```
codemercenary: Actually, this is behavior I want to support.  You should only throw an exception when the altitudes are equal.

autowiring/AutoPacket.h
veronicaz41: I wish compiler can also catch this ;)

autowiring/AutowiringDebug.h
veronicaz41: ok

autowiring/DecorationDisposition.h
veronicaz41: not really, see my reply below.

autowiring/DecorationDisposition.h
veronicaz41: There are two reasons I use `set` here:
1. prevent reduplicated introductions
2. more importantly, it also keeps `m_subscribers` sorted. according to
```
    bool operator<(const Subscriber& rhs) const {
      return std::tie(altitude, satCounter) > std::tie(rhs.altitude, rhs.satCounter);
    }
```
Otherwise, this test will fail. https://github.com/leapmotion/autowiring/pull/781/files#diff-6573071425e6af41616227f5dc6d7002R83

Not sure if there is a better way to guarantee these higher up. Suggestions?

veronicaz41:
@codemercenary ready for another pass

geminatea:
++ for this feature.  In-place modification in AFs is an essential feature.

codemercenary:
For better or for worse, merged.  I wonder how users will use this.

PR #782, opened 2015-10-15
Fix glaring mistake with autowiring::parallel void lambdas

codemercenary:
We correctly track the completion of these lambdas, but critically, _we don't actually call them_.  Fix this and add a test to ensure we don't make this mistake again.


veronicaz41:
:+1:

PR #783, opened 2015-10-15
Add `barrier` concept to autowiring::parallel

codemercenary:
There are cases, especially when dealing with lambdas that do not return anything, where the user doesn't care to iterate over each available item and just wants to return when everything currently outstanding is completed.


veronicaz41:
:shipit:

PR #786, opened 2015-10-15
Implement SetThreadPriority

### LINKED ISSUE #784: Implement SetThreadPriority on Unix
I'm tired of looking at the warning in PriorityBoost.cpp, it's about time we got around to implementing this.

It's likely all we need to do is to call the `nice` function or near equivalent, we just need to make sure we do so in a way that will also work properly on Mac.

codemercenary:
@gittyupagain implemented this functionality on a private branch awhile back, let's get it into mainline.

Fixes #784

### DIFF
--- a/src/autowiring/CoreThreadLinux.cpp
+++ b/src/autowiring/CoreThreadLinux.cpp
@@ -4,6 +4,7 @@
 #include "BasicThreadStateBlock.h"
 #include <pthread.h>
 #include <sys/resource.h>
+#include <pthread.h>
 
 using std::chrono::seconds;
 using std::chrono::milliseconds;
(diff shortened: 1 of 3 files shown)


src/autowiring/CoreThreadLinux.cpp
```diff
@@ -23,3 +24,39 @@ void BasicThread::GetThreadTimes(std::chrono::milliseconds& kernelTime, std::chr
...
+  case ThreadPriority::Highest:
+    percent = 83;
+    break;
+  case ThreadPriority::TimeCritical:
+    percent = 100;
+    break;
+  default:
+    throw std::runtime_error("Attempted to assign an unrecognized thread priority");
+  }
+  param.sched_priority = PTHREAD_MIN_PRIORITY + (percent*(PTHREAD_MAX_PRIORITY - PTHREAD_MIN_PRIORITY) + 50) / 100;
```
hham: #include <sched.h>
int sched_get_priority_max(int policy);
int sched_get_priority_min(int policy);
should be used here.

src/autowiring/CoreThreadMac.cpp
```diff
@@ -42,3 +51,38 @@ void BasicThread::GetThreadTimes(std::chrono::milliseconds& kernelTime, std::chr
...
+    break;
+  case ThreadPriority::TimeCritical:
+    percent = 100;
+    break;
+  default:
+    throw std::runtime_error("Attempted to assign an unrecognized thread priority");
+  }
+  param.sched_priority = PTHREAD_MIN_PRIORITY + (percent*(PTHREAD_MAX_PRIORITY - PTHREAD_MIN_PRIORITY) + 50) / 100;
+
+  pthread_setschedparam(m_thisThread.native_handle(), policy, &param);
```
hham: m_state-> is missing in front of m_thisThread.

src/autowiring/CoreThreadMac.cpp
```diff
@@ -11,8 +11,17 @@
...
 #include <iostream>
+#include <pthread.h>
 #include THREAD_HEADER
 
+// Missing definitions from pthreads.h
+#if !defined(PTHREAD_MIN_PRIORITY)
+#define PTHREAD_MIN_PRIORITY  0
+#endif
+#if !defined(PTHREAD_MAX_PRIORITY)
+#define PTHREAD_MAX_PRIORITY 31
```
hham: It looks like 63 is max priority in user space according to https://developer.apple.com/library/mac/documentation/Darwin/Conceptual/KernelProgramming/scheduler/scheduler.html.
Why did you put 31?

gittyupagain: This was done 20 months ago and never committed. It was an initial attempt at the problem. Those numbers probably came from Linux documentation. It is unfortunate that it isn't defined. If 63 is correct, then that should be used instead.

codemercenary: I'll make the change.

PR #790, opened 2015-10-16
Fix install command

codemercenary:
When a relative directory is given as the value to the DESTINATION argument of the cmake [`install`](https://cmake.org/cmake/help/v3.0/command/install.html) command, it is already treated as being relative to `CMAKE_INSTALL_PREFIX`.  When we redundantly specify it, then the PACKAGE operation can fail if the user specified `CMAKE_INSTALL_PREFIX` during configuration because it is interpreted to an absolute path at configuration time and is not resolved again when it has to be overridden at package time.


veronicaz41:
@codemercenary needs to resolve conlficts

PR #774, opened 2015-10-08
Fix issue with unlink teardown behavior

codemercenary:
This pull request highlights a problem with the way that unlink teardown occurs, specifically during signal unregistration.  Fix this problem.

- [ ] Fixes #762

PR #794, opened 2015-10-21
Make unit test more reliable, fix resulting failure

### LINKED ISSUE #791: AutoSignalTest.OuterPostDereference failing on Windows
Fix this so we can deploy

Here's the output from gtest:

```
C:\Development\autowiring\src\autowiring\test\AutoSignalTest.cpp(630): error: Value of: magic
  Actual: 15987178197214944733
Expected: 0xDEADBEEFC0FECAFE
Which is: 16045690984335330046
```

codemercenary:
This unit test was not catching a defect on Mac/Linux because of the way that the compiler can optionally optimize out write operations that occur in the destructor.  Change the test to make use of a static member instead so as to avoid this.

- [x] Fixes #791

### DIFF
--- a/autowiring/Autowired.h
+++ b/autowiring/Autowired.h
@@ -223,7 +223,7 @@ public:
       MemoEntry* entry = &context->FindByDeferrableAutowiring(this);
 
       autowiring::registration_t reg =
-        entry->m_sig += [this, fn] (autowiring::registration_t registration){
+        entry->m_sig += [this, fn] (autowiring::registration_t registration) mutable {
           fn();
           for(auto iter = m_autowired_notifications.begin(); iter != m_autowired_notifications.end(); ++iter) {
             if( *iter == registration) {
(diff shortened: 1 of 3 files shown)


codemercenary:
Ok, this test seems to pass on Mac, but fails on Windows.  Something is screwy here.

hham:
This passes on Linux, too.

codemercenary:
I just rebased to 18f49072c7f7781fc9694b72595c97a26bfdf0a4, which ensures this PR is building #774.  Still it passes.  I am going to try something else.

hham:
I still get the same error on Windows. In fact, outer(&OuterType::sig) += [this] {...} is not called on Mac and Linux because outer is already destroyed and not autowired any more. That's why this test passes on Mac and Linux.

codemercenary:
Good insight.  Let me see if I can make this test more reliable.

hham:
Looks great!!

PR #796, opened 2015-10-24
Address unsynchronized reset issue

codemercenary:
This pull request only partially addresses the asynchronous reset problem.  There is no general solution to this problem because at the heart of the issue is a cancellation race with `NotifyWhenAutowired`.  The simplest case that demonstrates this issue is as follows:

Thread 1:
```C++
{
    std::lock_guard<std::mutex> lk(lock);
    Autowired<T> ptr;
    ptr->NotifyWhenAutowired([&] {
        std::lock_guard<std::mutex> lk(lock);
        // ...do more work...
    });
    std::this_thread::sleep_for(std::chrono::seconds(1));
}
```

If `T` is injected during the sleep, the callback will be invoked and the injecting thread will block on the lock acquisition attempt.  The sleeping thread wakes up, and tries to destroy `ptr`.  This causes `DeferrableAutowiring::~DeferrableAutowiring()` to be called.

1.  If this destructor blocks until all registered handlers have been cancelled, a deadlock results.
2.  If it cancels on a best-effort basis, a potential use-after-free results.

The only solution is to require that users externally synchronize `Autowired::~Autowired` with `CoreContext::Inject`.

### DIFF
--- a/autowiring/CoreContext.h
+++ b/autowiring/CoreContext.h
@@ -37,7 +37,6 @@ class BoltBase;
 class CoreContext;
 class GlobalCoreContext;
 class JunctionBoxBase;
-class OutstandingCountTracker;
 
 template<typename T>
 class Autowired;
@@ -52,9 +51,6 @@ template<typename T>
 class JunctionBox;
 
 namespace autowiring {
-  template<typename... Args>
-  struct signal_relay_t;
-
   class ThreadPool;
 }
 
--- a/autowiring/CoreContext.h
+++ b/autowiring/CoreContext.h
@@ -3,7 +3,6 @@
 #include "AnySharedPointer.h"
 #include "auto_signal.h"
 #include "AutoFilterDescriptor.h"
-#include "AutowirableSlot.h"
 #include "AutowiringEvents.h"
 #include "autowiring_error.h"
 #include "Bolt.h"
@@ -36,13 +35,9 @@ struct CoreContextStateBlock;
 class BasicThread;
 class BoltBase;
 class CoreContext;
-class DeferrableAutowiring;
 class GlobalCoreContext;
 class JunctionBoxBase;
 class OutstandingCountTracker;
-
-template<class T, class Fn>
-class AutowirableSlotFn;
 
 template<typename T>
 class Autowired;
@@ -72,27 +67,6 @@ namespace autowiring {
 enum class ShutdownMode {
   Graceful,   ///< Shut down gracefully by allowing threads to run down dispatch queues.
   Immediate   ///< Shut down immediately, do not attempt to run down thread dispatch queues.
-};
-
-class AutoSearchLambda {
-public:
-  /// <summary>
-  /// Invoked at the conclusion of a FindByTypeRecursive search operation
-  /// </summary>
-  /// <remarks>
-  /// No information is passed about the success or failure of the search operation.  Implementors of this interface are
-  /// required to store enough context about the AnySharedPointer passed as the first argument to FindByTypeRecursive to
-  /// determine the status of this operation, if that status is interesting.
-  /// </remarks>
-  virtual void operator()() const = 0;
-};
-
-class AutoSearchLambdaDefault:
-  public AutoSearchLambda
-{
-public:
-  // AutoSearchLambda overrides:
-  void operator()() const override {}
 };
 
 /// <summary>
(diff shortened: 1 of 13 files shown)


autowiring/CoreContext.h
```diff
@@ -1137,75 +1137,6 @@ class CoreContext:
...
-
-  /// <summary>
-  /// Registers a slot to be autowired
-  /// </summary>
-  /// <param name="ref">The space where the resolved shared pointer will be stored</param>
-  /// <param name="defer">
-  /// In the event that resolution was not successful, the deferrable that will be used to perform delayed satisfaction
-  /// </param>
-  template<class T>
-  void Autowire(AnySharedPointerT<T>& ref, DeferrableAutowiring& defer) {
```
jdonald: This function is still called on line 105 of autowiring/Autowired.h, leading to a build error on all systems. Removing that clause reveals other compilation issues.

codemercenary: This PR is still underway.

jdonald: Sorry about that. I got too excited and jumped the gun on attempting to test this w/ unit tests/DumpReplay.

autowiring/CoreContext.h
```diff
@@ -670,6 +622,16 @@ class CoreContext:
...
   }
 
+  /// \internal
+  /// <summary>
+  /// Scans the memo collection for the specified entry, or adds a deferred resolution marker if resolution was not possible
+  /// </summary>
+  /// <returns>
+  /// The memo entry where this type was found
+  /// </returns>
+  /// <param name="id">The type to be located</param>
```
veronicaz41: Should be `<param name="type">`

src/autowiring/CoreContext.cpp
```diff
@@ -797,8 +795,7 @@ void CoreContext::UnregisterEventReceiversUnsafe(void) {
     m_pParent->RemoveEventReceivers(m_eventReceivers);
 
   // Recursively unregister packet factory subscribers:
-  AnySharedPointerT<AutoPacketFactory> pf;
-  FindByTypeUnsafe(pf);
+  FindByTypeUnsafe(auto_id_t<AutoPacketFactory>{});
```
veronicaz41: How does this unregister packet factory subscribers?

codemercenary: Good freaking question...let me review this.

codemercenary: It doesn't.  This is handled in `AutoPacketFactory::OnStop`

src/autowiring/AutowirableSlot.cpp
```diff
@@ -8,45 +8,73 @@
...
+  // that we perform an expensive memory reallocation, so let's pay a little memory in advance to help make the
+  // most common use case pretty fast.
+  m_autowired_notifications.reserve(2);
+  m_autowired_notifications.emplace_back(std::move(reg));
 }
 
-void DeferrableAutowiring::RegisterDeferredAutowire(autowiring::registration_t&& reg) {
-  m_deferred_registration = std::move(reg);
+DeferrableAutowiring::~DeferrableAutowiring(void) {
+  // BUG:  The reset call does not strictly prohibit
```
veronicaz41: Why this is a bug?

codemercenary: Forgot to explain this fully.

src/autowiring/AutowirableSlot.cpp
```diff
@@ -8,45 +8,80 @@
...
+  // Local versions of the members we are clearing out
+  std::weak_ptr<CoreContext> contextWeak;
+  AnySharedPointer ptr;
+  std::vector<autowiring::registration_t> autowired_notifications;
+  
+  // Move and handle resources under lock
+  {
+    std::lock_guard<autowiring::spin_lock> lk{ m_lock };
+    autowired_notifications = std::move(m_autowired_notifications);
+    ptr = std::move(m_ptr);
```
veronicaz41: Why we need to move `m_ptr` here?

veronicaz41: OK. this is to make sure `m_ptr` is released.

src/autowiring/CoreContext.cpp
```diff
@@ -300,81 +301,78 @@ void CoreContext::AddInternal(const AnySharedPointer& ptr) {
...
     throw autowiring_error("This interface is already present in the context");
 
   // Now we can satisfy it:
+  entry.m_local = true;
   entry.m_value = ptr;
   UpdateDeferredElement(std::move(lk), entry);
 }
 
-MemoEntry& CoreContext::FindByType(AnySharedPointer& reference, bool localOnly) const {
+MemoEntry& CoreContext::FindByType(auto_id type, bool nonrecursive) const {
```
jdonald: This breaks compatibility with an existing use case.

    std::shared_ptr<WinUSBPayloadReader> rdr;
    streaming->FindByType<WinUSBPayloadReader>(rdr);

Given we can't use auto_id there, what is the replacement? Also, would this push the version to 0.8.0?

codemercenary: I'll add a variant that will restore compatibility.

codemercenary: Are you sure this breaks compatibility?  `CoreContext.h:1107` provides the shared pointer version of this routine.

PR #804, opened 2015-10-27
Eliminate unused type

codemercenary:
This is a relic of the old AutoConfig system, should have been removed when that was.


jnguyen75:
:(

codemercenary:
I know.  It makes me sad too.

PR #808, opened 2015-10-27
Fix busted FindByType override

codemercenary:
This method got broken in #796, correct the problem and improve test coverage.


jdonald:
Verified this in platform via SystemTest with WinUSB (which uses FindByType).

PR #814, opened 2015-10-28
Split out atomic linked list behaviors into their own datatype

codemercenary:
There are some signal issues that are related to bugs with an atomic list behavior on Linux.  Create a dedicated atomic linked list concept that allows us to build and test a reliable atomic linked list concept separate from the rest of signal.

- [x] Ensure this type can be used with polymorphic types without needing reallocation


codemercenary:
This test it's failing on all platforms, I think it is the root cause of our signal instability.  Will investigate when I get home.

PR #826, opened 2015-11-17
Fix incorrect handling of private ctors

codemercenary:
In cases where a factory new method is available on a type, we should not even be attempting to invoke that type's constructor, because the constructor may be private or might not even exist.

Eliminate any instantiations of templates that attempt such invocation.

- [x] Compiles on Windows


veronicaz41:
@codemercenary need to resolve conflicts

codemercenary:
Done.

PR #827, opened 2015-11-19
Fix issue with post-hoc NotifyWhenAutowired invocation

codemercenary:
Looks like the current context is not being properly assigned in this case, add a test to prove this and then fix the defect.


veronicaz41:
@codemercenary need to resolve conflicts

PR #833, opened 2015-12-08
Add missing newlines at the end of source files

jdonald:
Should clean these up in autowiring too, or else msbuild.exe warns about these when including some autowiring headers.


hham:
Looks good.

PR #837, opened 2015-12-19
Add extended information about null field dereferences

### LINKED ISSUE #835: More verbosity with "Attempted to dereference a null autowired field"
I get this error frequently, but the message is very unhelpful. Surely with RTTI we can at least print out a string representation of the offending type?

codemercenary:
Fixes #835, just adds a little bit more information about null dereference exceptions when they occur.


jhe-:
Exactly what I needed. Thanks for the speedy response!

codemercenary:
My pleasure.

PR #840, opened 2016-01-06
Fix toolchain file again

codemercenary:
Eliminate the `ARMv8` flag completely, just use different toolchain files for 32 and 64 bit.  The two require different compilers anyway, makes sense to divide up the toolchain files as well.  Furthermore, we no longer need `#include <asm/page.h>` on our target build platforms, so we can remove this.


hham:
+1 for separating toolchain file. I am testing this now.

hham:
Works fine.

jdonald:
:+1:

PR #846, opened 2016-02-09
Implement a try-to-dispatch concept

codemercenary:
This is useful for cases where consumers want to be able to retry lambdas from the dispatch queue while preserving order in single-threaded use cases.

### DIFF
--- a/src/autowiring/DispatchQueue.cpp
+++ b/src/autowiring/DispatchQueue.cpp
@@ -67,6 +67,31 @@ void DispatchQueue::DispatchEventUnsafe(std::unique_lock<std::mutex>& lk) {
   (*thunk)();
 }
 
+void DispatchQueue::TryDispatchEventUnsafe(std::unique_lock<std::mutex>& lk) {
+  // Pull the ready thunk off of the front of the queue and pop it while we hold the lock.
+  // Then, we will excecute the call while the lock has been released so we do not create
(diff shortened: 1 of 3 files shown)


src/autowiring/DispatchQueue.cpp
```diff
@@ -67,6 +67,31 @@ void DispatchQueue::DispatchEventUnsafe(std::unique_lock<std::mutex>& lk) {
...
+  // deadlocks.
+  DispatchThunkBase* pThunk = m_pHead;
+  m_pHead = pThunk->m_pFlink;
+  lk.unlock();
+
+  try { (*pThunk)(); }
+  catch (...) {
+    // Failed to execute thunk, put it back
+    lk.lock();
+    pThunk->m_pFlink = m_pHead;
```
veronicaz41: From reading the comments in the header, I thought we are putting this thunk back to the tail of the queue?

veronicaz41:
Interesting. What is the use case for this?

src/autowiring/DispatchQueue.cpp
codemercenary: Negative, comment states that the item is replaced at the front of the queue on failure.

veronicaz41: I see.

PR #851, opened 2016-02-25
Rename auto_signal.h to signal.h

codemercenary:
The type included here is called signal, the header should have the same name.  Also preserve the auto_signal header file for legacy customers, just have it punt over to the newly named file.


hham:
Good to go.

PR #854, opened 2016-02-26
Allow callers to detect `DispatchQueue` drops

codemercenary:
The return value of `DispatchQueue::operator+=` should signify whether the lambda was pended to the queue or not.  Right now it's void, so it's trivial to just return a boolean here instead.


hham:
Looks good.

PR #860, opened 2016-03-11
Add a tweakable concept

codemercenary:
Tweakable is different from config in that the modified value is also the accessed value.  Only atomic types can be considered here as a result, because asynchronous modification is possible, and therefore these members are necessarily considered `volatile`.

### DIFF
--- a/src/autowiring/tweak.h
+++ b/src/autowiring/tweak.h
@@ -1,3 +1,4 @@
+// Copyright (C) 2012-2015 Leap Motion, Inc. All rights reserved.
 #pragma once
 #include <atomic>
 #include <type_traits>
--- a/src/autowiring/tweak.h
+++ b/src/autowiring/tweak.h
@@ -59,6 +59,9 @@ namespace autowiring {
     operator T(void) volatile const {
       return value;
     }
+    volatile T* operator&(void) volatile {
+      return &value;
+    }
   };
 
   template<typename T>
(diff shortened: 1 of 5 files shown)


GabrielHare:
@codemercenary build failed for clang 6.0

src/autowiring/tweak.h
```diff
@@ -0,0 +1,87 @@
+// Copyright (C) 2012-2015 Leap Motion, Inc. All rights reserved.
+#pragma once
+#include <atomic>
+#include <type_traits>
+
+namespace autowiring {
+  template<typename T>
+  struct tweak {
+    static_assert(std::is_arithmetic<T>::value, "Cannot define a tweakable value that is not arithmetic");
```
GabrielHare: Are arithmetic types guaranteed to be modified atomically?

src/autowiring/tweak.h
```diff
@@ -0,0 +1,87 @@
...
+    /// </summary>
+    /// <returns>
+    /// True if this instance was dirty
+    /// </returns>
+    bool clear_dirty(void) volatile {
+      if(value == last_observed)
+        return false;
+      last_observed = value;
+      return true;
+    }
```
GabrielHare: What is the intended use of "is_dirty()" and "clear_dirty()"? Since value is both assigned to and returned, it appears to mat that there is no guarantee in a multi-threaded application that the state of a tweak will be clean following a call to "clear_dirty".

What am I missing? And, can you write a unit test to demonstrate the utility of these methods?

codemercenary:
Closing this PR as superceded

PR #863, opened 2016-03-16
Retroactive bolting for existing contexts

codemercenary:
Historically, bolts are only asserted for _new_ contexts which match the bolt pattern.  This should be changed so that a bolt will be asserted for any context which already exists at the time the bolt is introduced.

When a bolt is added to a context, it should be asserted for all parents and children that are already created which satisfy the bolt requirement.  This allows bolts to be added potentially at some point after a context is created and running, and allows a new order-independent pattern to be considered the standard behavior in broader parts of the system.

Also change the bolt system so that it internally uses `auto_id` instead, this type is faster and more flexible than the STL `type_info`, and fix the hashing routine so that `auto_id{}` may be hashed.

### DIFF
--- a/src/autowiring/BoltBase.h
+++ b/src/autowiring/BoltBase.h
@@ -1,10 +1,8 @@
 // Copyright (C) 2012-2015 Leap Motion, Inc. All rights reserved.
 #pragma once
-#include <typeinfo>
+#include "auto_id.h"
 
 class CoreContext;
-
-typedef const std::type_info*const* t_TypeInfoVector;
 
 /// <summary>
 /// </summary>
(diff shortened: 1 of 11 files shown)


src/autowiring/BoltBase.h
```diff
@@ -1,11 +1,9 @@
...
-#include <typeinfo>
+#include "auto_id.h"
 
 class CoreContext;
 
-typedef const std::type_info*const* t_TypeInfoVector;
-
 /// <summary>
 /// </summary>
 /// <param name="contextName">The name of the context whose creation we want to listen for</param>
```
veronicaz41: This comment should be removed, no "contextName" here

PR #866, opened 2016-03-23
Fix transitive unsatisfiability defect

codemercenary:
There is an issue where a filter producing an output will not have its outputs marked unsatisfiable if some of that filter's required input arguments are themselves unsatisfiable.  Add a test and correct the issue by implementing a transitive unsatisfiability scan.


src/autowiring/test/AutoFilterSatisfiabilityTest.cpp
```diff
@@ -0,0 +1,63 @@
+// Copyright (C) 2012-2015 Leap Motion, Inc. All rights reserved.
+#include "stdafx.h"
+#include <autowiring/autowiring.h>
+#include "TestFixtures/Decoration.hpp"
+
+class SatisfiabilityTest:
```
veronicaz41: Should probably also change the class to `AutoFilterSatisfiabilityTest`

codemercenary: :+1:

PR #869, opened 2016-03-31
Add rvalue shared pointer support to AutoFilter

veronicaz41:
Add support for rvalue shared pointer arguments for AutoFilter. It would make it possible to replace or remove a decoration in place.

### DIFF
--- a/src/autowiring/AutoPacket.cpp
+++ b/src/autowiring/AutoPacket.cpp
@@ -114,7 +114,7 @@ void AutoPacket::AddSatCounterUnsafe(SatCounter& satCounter) {
       // otherwise insert it to the right position so that the modifiers vector is sorted by altitude
       auto it = entry.m_modifiers.begin();
       while (it != entry.m_modifiers.end()) {
-        if (it->altitude == satCounter.GetAltitude() && it->is_shared == pCur->is_shared) {
+        if (it->altitude == satCounter.GetAltitude()) {
           std::stringstream ss;
           ss << "Added multiple rvalue decorations with same altitudes for type " << autowiring::demangle(pCur->id);
           throw autowiring_error(ss.str());
--- a/src/autowiring/AutoPacket.cpp
+++ b/src/autowiring/AutoPacket.cpp
@@ -114,7 +114,7 @@ void AutoPacket::AddSatCounterUnsafe(SatCounter& satCounter) {
       // otherwise insert it to the right position so that the modifiers vector is sorted by altitude
       auto it = entry.m_modifiers.begin();
       while (it != entry.m_modifiers.end()) {
-        if (it->altitude == satCounter.GetAltitude()) {
+        if (it->altitude == satCounter.GetAltitude() && it->is_shared == pCur->is_shared) {
           std::stringstream ss;
           ss << "Added multiple rvalue decorations with same altitudes for type " << autowiring::demangle(pCur->id);
           throw autowiring_error(ss.str());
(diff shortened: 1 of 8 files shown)


veronicaz41:
@codemercenary ready for review

src/autowiring/AutoPacket.cpp
```diff
@@ -274,6 +271,48 @@ void AutoPacket::UpdateSatisfactionUnsafe(std::unique_lock<std::mutex> lk, const
...
+    }
+  };
+
+  for (auto modifier : disposition.m_modifiers) {
+    if (!modifier.satCounter)
+      continue;
+    auto& satCounter = *modifier.satCounter;
+    if (modifier.is_shared) {
+      if (disposition.m_decorations.size() > 1)
+        throw autowiring_error("An AutoFilter was detected which has single-decorate rvalue shared pointer argument in a graph with multi-decorate outputs");
```
codemercenary: Is there any way we can detect this case sooner?

src/autowiring/test/AutoFilterRvalueTest.cpp
```diff
@@ -95,44 +88,51 @@ TEST_F(AutoFilterRvalueTest, MultipleModifiersWithSameAltitude) {
 }
 
 TEST_F(AutoFilterRvalueTest, MultipleModifiersWithDifferentAltitude) {
-  AutoRequired<AutoPacketFactory> factory;
   int called = 0;
+  int order[] = {-1, -1, -1, -1, -1, -1};
```
codemercenary: Good call on this test modification

src/autowiring/AutoPacket.cpp
veronicaz41: The tricky thing here is because shared pointer modifier can do things like `sp.reset()`, `sp.reset(new Decoration<1>)`. It can change the number of decorations (`disposition.m_decorations.size()`).
That is why when we found out an modifier `is_shared`, we call it right there (along with everything else already in the `callQueue`) rather than put it in the `callQueue`. And then check `disposition.m_decorations.size()` inside the loop, because it may change after the call to shared pointer modifier.

veronicaz41: On the second thought, the shared pointer modifier can only change the number of decorations from 0 to1 or from 1 to 0. So we should be able to detect this case a little sooner, like above the loop.

src/autowiring/AutoPacket.cpp
```diff
@@ -274,6 +271,50 @@ void AutoPacket::UpdateSatisfactionUnsafe(std::unique_lock<std::mutex> lk, const
...
+    for (size_t i = satCounter.GetArity(); i--;) {
+      // Only consider output arguments:
+      if (args[i].is_output)
+        // This output is transitively unsatisfiable, include it for later removal
+        unsatOutputArgs.push_back(&args[i]);
+    }
+  };
+
+  if (!disposition.m_modifiers.empty() && disposition.m_decorations.size() > 1)
+    throw autowiring_error("An AutoFilter was detected which has single-decorate rvalue argument in a graph with multi-decorate outputs");
```
codemercenary: What I mean by "earlier", here, is that I would like to know if it's possible to check for this condition right when the packet is created, as opposed to here, when the decoration is being added.

codemercenary: Or possibly we could detect this condition on the `AutoPacketFactory` itself, and then only check it again when this packet has a lambda attached to it.

veronicaz41: Added the checking to `AddSatCounterUnsafe`. It will be called when the packet is created and also when the packet has a lambda attached to it.
For our current implementation, don't think it is possible to detect this condition on the AutoPacketFactory itself, because when adding an autofilter, all it does is insert it to a linked list. We only construct the graph when a new packet is created.

codemercenary: That should be fine, then.  I'll merge this now.

PR #870, opened 2016-03-31
Eliminate stray "orig" file

codemercenary:
Got into develop awhile back, get rid of it


veronicaz41:
Should probably add `*.orig` to .gitignore

PR #872, opened 2016-04-01
future is only supported by android64 gcc 4.9

hham:
Unfortunately, Android32 gcc 4.9 does not support future.
This is required to upgrade android32 gcc4.8 to gcc 4.9.

### DIFF
--- a/src/autowiring/C++11/cpp11.h
+++ b/src/autowiring/C++11/cpp11.h
@@ -127,7 +127,7 @@
 /*********************
  * future availability
  *********************/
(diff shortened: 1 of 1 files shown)


codemercenary:
:+1:

jdonald:
This logic is a bit different from what we have in platform. Ideally the limitation will get fixed on Android 32-bit with gcc 5, yet the logic here would continue using boost/future for all 32-bit Android.

However, I haven't confirmed that it'll get fixed, and up until recently thought the same about gcc 4.9. It's also likely that we'll switch over to the clang ndk rather than gcc 5, so we can deal with that when the time comes.

hham:
Agree. In fact, I wasn't sure that gcc 5.0 on android 32bit will includes future. I will keep it in mind to fix this when the time comes.

PR #878, opened 2016-04-11
Eliminate use of `std::to_string`

codemercenary:
This method isn't supported in Android.  Remove our use of it until a shim is created, or support is provided.


jdonald:
Discussed w/ Jason to understand why we can't easily use `template <typename T>
  std::string to_string` inside autowiring, as it would conflict with platform's use of the same.

PR #882, opened 2016-04-11
Fix Android build

codemercenary:
Do not specify the current working directory as an include path, this jacks up the android build due to our use of the "signal.h" filename.


hham:
I get the following compile error.

/home/ubuntu/Projects/autowiring/autowiring/contrib/autoboost/libs/filesystem/portability.cpp:14:20: fatal error: stdafx.h: No such file or directory
 #include "stdafx.h"

codemercenary:
@hham should be fixed now.

codemercenary:
Actually hold on this, there is one more thing I need to fix.

codemercenary:
Great.  I didn't mean to merge this one either.

PR #888, opened 2016-04-13
Add nullptr_t overload to AnySharedPointer

codemercenary:
Should be allowed, sometimes there are runtime settings where `nullptr_t` is desired.

### DIFF
--- a/src/autowiring/auto_id.h
+++ b/src/autowiring/auto_id.h
@@ -99,17 +99,44 @@ namespace autowiring {
   };
 }
 
+struct auto_id;
+
+/// <summary>
+/// Identifier sigil structure
+/// </summary>
+/// <remarks>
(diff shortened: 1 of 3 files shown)


GabrielHare:
Looks good. I'd suggest adding construct & compare unit tests to ensure continued support of this feature.

src/autowiring/auto_id.h
```diff
@@ -99,17 +99,44 @@ namespace autowiring {
...
+    auto_id(auto_id_t<T>::s_block)
+  {}
+
+  auto_id(const autowiring::auto_id_block& block) :
+    block(&block)
   {}
 
   // A pointer to the identifier block containing the index
-  const autowiring::auto_id_block* block = nullptr;
+  const autowiring::auto_id_block* block;
```
GabrielHare: Why is no initialization better?

PR #890, opened 2016-04-13
Eliminate TypeIdentifier

codemercenary:
This type should have been removed with AutoFired, now we can finish the job.


codemercenary:
Neeeeevermind.  `TypeRegistry` still uses this.

PR #895, opened 2016-04-15
Add support for void AnySharedPointer

codemercenary:
Allow users to retrieve `std::shared_ptr<void>` directly from an `AnySharedPointer`.

### DIFF
--- a/src/autowiring/AnySharedPointer.h
+++ b/src/autowiring/AnySharedPointer.h
@@ -68,6 +68,8 @@ public:
     return obj && try_assign(obj);
   }
 
+  const std::shared_ptr<void>& as_void(void) const { return m_ptr; }
+
   template<class T>
   const std::shared_ptr<T>& as(void) const {
     // The safety of this routine is verified by the AnySharedPointer unit tests
--- a/src/autowiring/test/AnySharedPointerTest.cpp
+++ b/src/autowiring/test/AnySharedPointerTest.cpp
@@ -267,3 +267,9 @@ TEST_F(AnySharedPointerTest, NullPtrConstruction) {
   ASSERT_EQ(auto_id_t<void>{}, y.type());
   ASSERT_EQ(auto_id_t<void>{}, z.type());
 }
+
+TEST_F(AnySharedPointerTest, VoidSharedPointer) {
+  auto p = std::make_shared<int>(101);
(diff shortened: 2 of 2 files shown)


src/autowiring/AnySharedPointer.h
```diff
@@ -68,6 +68,8 @@ struct AnySharedPointer {
     return obj && try_assign(obj);
   }
 
+  const std::shared_ptr<void>& as_void(void) const { return m_ptr; }
+
```
ajohnston33: This looks like a means of having an un-typed container for a pointer without losing type safety.  I can understand wanting this to be able to write simpler template logic elsewhere.  What confuses me here is throwing away your ability to guard type safe access by returning a shared_ptr<void>.  Shouldn't callers need to provide the type of the underlying pointer so it can be checked before being allowed access to the shared pointer?

codemercenary: Quite correct.  Actually, the main motivation to using `AnySharedPointer` is not necessarily that the code becomes any simpler, but that I can avoid using templated functions altogether and move to runtime versions of the same.  These runtime versions are compiled once and don't increase build times downstream.

This particular routine is mainly intended to make unit tests a little easier to write in that sometimes I don't actually care about the underlying type, I just want whatever the pointed-to object is.  There are implicit dangers of doing this--in particular:

```C++
std::shared_ptr<Derived> p;
AnySharedPointer x = std::static_pointer_cast<Base>(p);
ASSERT_EQ(p.get(), x.as_void());  // Might fail!
```

codemercenary: Also when I use this type I get burned almost every time and have to rely pretty heavily on unit tests to ensure correctness.  It's definitely not the friendliest thing in the world.

ajohnston33: I am thinking something like this?
```template<class T>
std::shared_ptr<T> as_shared(void) const {
  "Check that auto_id of T == m_ti"
  return static_pointer_cast<T>(m_ptr);
}```

codemercenary: In fact, I have something like that now, it's called [`as`](https://github.com/leapmotion/autowiring/pull/895/files?diff=unified#diff-5de4567ee35f6903d932573f2bc0c4b2R74)

PR #898, opened 2016-04-15
Move post-construction after everything

codemercenary:
`AutoInit` post-construction should occur after injection is guaranteed, so that for a particular context and type, `AutoInit` is called exactly once.  If this invocation of `AutoInit` throws an exception, the context should be shut down.


src/autowiring/test/PostConstructTest.cpp
```diff
@@ -427,3 +414,19 @@ TEST_F(PostConstructTest, CorrectContextAssignment) {
 
   ASSERT_EQ(child, currentCtxt) << "Current context was not correctly assigned in a post-satisfied NotifyWhenAutowired handler";
 }
+
+namespace {
+  class ThrowsInAutoInit {
+  public:
+    void AutoInit(void) {
+      throw std::runtime_error("Failure!");
```
ajohnston33: Consider a message that is a unique string users can locate more easily.  E.g. "AutoInit failure"

PR #901, opened 2016-04-18
Add some operator overload support for comparison to nullptr_t

codemercenary:
Some clients prefer to explicitly test `Autowired<T>` against `nullptr`, and this was supported in prior releases.  Provide the necessary operator overloads to do so.


wal-rus:
+1, has tests, and these are pretty useful.

PR #902, opened 2016-04-18
Updated boost::config and boost::asio to boost 1.60 for msvc2015

wal-rus:
This is boost::config is required for MSVC2015 support, the update to boost::asio is for bugfixes


codemercenary:
Holy god.  Ok then, time to update.

PR #903, opened 2016-04-18
Update websocketpp to 0.7.0

wal-rus:
This is required for MSVC2015 compatibility. Note that we're jumping from 0.4.0, so some minor breaking changes have been introduced. We should verify that this doesn't break platform.

- [x] Merge #902


codemercenary:
Been meaning to do this for awhile.  Thanks.

PR #905, opened 2016-04-18
Eliminate MSVC warnings in websocketpp

codemercenary:
These warnings arose due to the upgrade to the latest websocketpp, fix them.

### DIFF
--- a/contrib/websocketpp/websocketpp/processors/hybi00.hpp
+++ b/contrib/websocketpp/websocketpp/processors/hybi00.hpp
@@ -122,7 +122,7 @@ public:
 
         res.append_header(
             "Sec-WebSocket-Key3",
-            md5::md5_hash_string(std::string(key_final.data(),16))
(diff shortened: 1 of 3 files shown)


contrib/websocketpp/websocketpp/processors/hybi00.hpp
```diff
@@ -114,13 +114,15 @@ class hybi00 : public processor<config> {
...
-                  &key_final[8]);
+        std::copy_n(
+            key3.begin(),
+            std::min<size_t>(8, key3.size()),
+            key_final.begin() + 8
+        );
 
         res.append_header(
             "Sec-WebSocket-Key3",
-            md5::md5_hash_string(std::string(key_final,16))
```
wal-rus: if you're using std::array, let's get rid of *all* the magic numbers eh?

codemercenary: Sure, why not.

PR #907, opened 2016-04-20
Fix whitespace and leftover un-renamed references in AutoBoost

wal-rus:
Adds a script that can be used to automatically generate AutoBoost from Boost, and fixes the existing version's trailing whitespace.


codemercenary:
:+1:

PR #908, opened 2016-04-20
Fix incorrect cast

codemercenary:
Careless mistake.  One of the consequences of not having sufficient testing of a contrib library.


wal-rus:
Websocketpp has a test suite - perhaps I could run it?

codemercenary:
Might be wise.

PR #913, opened 2016-04-24
Add support for VS2015

codemercenary:
The latest standard uses `wchar_t` as the underlying type for `std::filesystem::path`, and deprecates `std::filesystem::wpath`.  Update the shim header to reflect this change.


hham:
Ok to build.

PR #914, opened 2016-04-24
Fix Await overload return type.

codemercenary:
The return type for Await, here, is supposed to be a reference, because it's used in a byref sense by the caller.


hham:
Looks good.

PR #915, opened 2016-04-24
Eliminate invalid throw from destructor

codemercenary:
This has been here for a long time, but is actually incorrect behavior, and a throw here would result in an invocation of `std::terminate`.


jhe-:
Yeah, I also don't see what benefit you would get from throwing there.

PR #916, opened 2016-04-24
Add explicit operator bool overload

codemercenary:
This is intended to silence performance warnings in VS2015.  Also make a change to the `ASSERT_TRUE` and `ASSERT_FALSE` macros in order to eliminate the performance warning.


hham:
good to go.

PR #918, opened 2016-04-26
Fix VS2013 filesystem shim header problems

codemercenary:
Add a missing overload, fix missing return statement in filesystem shim header for VS2013

### DIFF
--- a/src/autowiring/C++11/filesystem.h
+++ b/src/autowiring/C++11/filesystem.h
@@ -27,6 +27,9 @@ namespace std {
       path(const string_type& _Str) :
         awfsnamespace::wpath(_Str)
       {}
+      path(const awfsnamespace::wpath& _Right) :
+        awfsnamespace::wpath(_Right)
+      {}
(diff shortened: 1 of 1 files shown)


src/autowiring/C++11/filesystem.h
```diff
@@ -38,7 +41,7 @@ namespace std {
       {}
 
       basic_path& operator=(basic_path&& _Right) { *(awfsnamespace::wpath*)this = std::move(_Right); }
-      basic_path& operator=(const string_type& _Str) { *(awfsnamespace::wpath*)this = _Str; }
+      basic_path& operator=(const string_type& _Str) { return *(awfsnamespace::wpath*)this = _Str; }
       basic_path& operator=(const wchar_t* _Ptr) { *(awfsnamespace::wpath*)this = _Ptr; }
```
veronicaz41: should those all have return?

codemercenary: Crap.  Yes they should.

Issue #920, opened 2016-05-04
Double-decoration error in Autowiring v1.0.0


codemercenary:
There's a defect introduced in #866 which is causing a multi-decorate problem to appear on initialization in certain user build cases.  This defect is serious enough to merit a patch release this week to correct the problem.

The issue is only reproduceable in platform.  This is a copy of the error log:

```
[Critical] Unhandled exception: Cannot decorate this packet with type SubFrame, the requested decoration is already satisfied
[Critical] An unhandled exception occurred in a device context: Cannot decorate this packet with type SubFrame, the requested decoration is already satisfied
```

veronicaz41:
This is related to this PR: https://github.com/leapmotion/autowiring/pull/866
And in platform, the exception throw at "ImageDebug"

PR #923, opened 2016-05-05
Fix double-decoration error

veronicaz41:
The problem is if an auto filter has multiple inputs marked as unsatisfiable, each time it is marked as unsatisfiable, we call `IncProducerCount`, so the auto filter will become complete prematurely.
The fix is to make sure `IncProducerCount` only got called once for each auto filter, only when all its inputs are satisfied or marked as unsatisfiable.

- [x] fixes https://github.com/leapmotion/autowiring/issues/920

### DIFF
--- a/src/autowiring/AutoPacket.cpp
+++ b/src/autowiring/AutoPacket.cpp
@@ -279,9 +279,11 @@ void AutoPacket::UpdateSatisfactionUnsafe(std::unique_lock<std::mutex> lk, const
 
   // Recursively mark unsatisfiable any single-output arguments on these subscribers:
   std::vector<const AutoFilterArgument*> unsatOutputArgs;
-  auto MarkOutputsUnsat = [&unsatOutputArgs] (SatCounter satCounter) {
-    if (!satCounter.Decrement())
+  auto MarkOutputsUnsat = [&unsatOutputArgs] (SatCounter& satCounter) {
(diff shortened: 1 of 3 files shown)


src/autowiring/AutoPacket.cpp
```diff
@@ -279,7 +279,9 @@ void AutoPacket::UpdateSatisfactionUnsafe(std::unique_lock<std::mutex> lk, const
 
   // Recursively mark unsatisfiable any single-output arguments on these subscribers:
   std::vector<const AutoFilterArgument*> unsatOutputArgs;
-  auto MarkOutputsUnsat = [&unsatOutputArgs] (const SatCounter& satCounter) {
+  auto MarkOutputsUnsat = [&unsatOutputArgs] (SatCounter satCounter) {
```
codemercenary: Why a copy?

veronicaz41: oh, a typo

src/autowiring/SatCounter.h
```diff
@@ -34,7 +34,7 @@ struct SatCounter:
   /// </summary>
   /// <returns>True if this decrement yielded satisfaction of all arguments</returns>
   bool Decrement(void) {
-    return !--remaining;
+    return remaining && !--remaining;
```
codemercenary: Nice, I like compact logic.

PR #924, opened 2016-05-05
Fix AutoNet regression

### LINKED ISSUE #922: Fix AutoNet
AutoNet seems to be having some pretty serious problems as of v1.0.0.  In particular, not all of the context information is correctly relayed to the front end, and it does not update continuously.

One possible problem is the recent update to websocketpp, another possibility is a change to how Autowiring reports context events to clients.  Investigate, and fix the problem.

codemercenary:
This was not caused by the websocketpp update; rather, it was caused when CoreContext made a change to the way context state is constructed when `AutoFired` was deprecated.  Fixing this just involves moving the enumeration responsibility to AutoNet, where it probably belonged all along.

Fixes #922


wal-rus:
Thanks for jumping on this one

PR #853, opened 2016-02-26
Make `SetThreadPriority` protected

codemercenary:
The user knows what they're doing.  Trust them.

PR #927, opened 2016-05-06
Fix `SetThreadPriority` comment

codemercenary:
The constraint described by this comment is no longer valid as of #853, fix it, and also allow users to set the thread priority before the thread is ever initiated.

### DIFF
--- a/src/autowiring/BasicThread.h
+++ b/src/autowiring/BasicThread.h
@@ -146,8 +146,7 @@ protected:
   /// Sets the thread priority of this thread
   /// </summary>
   /// <remarks>
-  /// This method must only be called from the running thread's context, and then, only inside
-  /// the ElevatePriority constructor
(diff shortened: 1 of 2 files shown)


src/autowiring/BasicThread.h
```diff
@@ -146,8 +146,7 @@ class BasicThread:
   /// Sets the thread priority of this thread
   /// </summary>
   /// <remarks>
-  /// This method must only be called from the running thread's context, and then, only inside
-  /// the ElevatePriority constructor
+  /// This method can only be safely called after the thread is already running.  Calling it before
```
gittyupagain: The comment doesn't appear to be complete.

codemercenary: That's what I get for trying to submit a PR on Cinco de Mayo.

PR #934, opened 2016-05-07
Correct unsynchronized reset

codemercenary:
The next packet shared pointer must be cleared outside of the lock in the same way as the collection of AutoFilters.  Store it in a local variable, nullify the member under lock, and destroy on scope exit.

Additionally, the stop check must occur under lock in order to guard against an unsynchronized state change.

This defect was introduced in #845


veronicaz41:
👍

PR #938, opened 2016-05-09
Add tid to BasicThread, make debug lists more readable.

codemercenary:
Another slight adjustment to hung runnable outputs.  Contexts should print all of their runnable members, whether they are running or not, and use a status flag to indicate what is still running and under what context.  Also print the thread ID of runnables that are threads, wherever possible.

Furthermore, the context map was not actually readable when printed in windows due to Unicode translation problems on the command line.  Replace the unicode box drawing code with simple ASCII in order to ensure compatibility for now, and update unit tests as needed.

Also add a unit test for all new behaviors.

### DIFF
--- a/src/autowiring/test/AutowiringDebugTest.cpp
+++ b/src/autowiring/test/AutowiringDebugTest.cpp
@@ -5,6 +5,7 @@
 #include "TestFixtures/Decoration.hpp"
 #include <algorithm>
 #include <fstream>
+#include <thread>
 
 class AutowiringDebugTest:
   public testing::Test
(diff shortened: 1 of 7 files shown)


src/autowiring/test/AutowiringDebugTest.cpp
```diff
@@ -56,13 +57,13 @@ TEST_F(AutowiringDebugTest, ContextPrintout) {
...
-                     "└── Herp\n"
-                     "    ├── Derp\n"
-                     "    └── Derp\n";
+                     "|-- Derp\n"
+                     "|   |-- Herp\n"
+                     "|   |   *-- int\n"
+                     "|   *-- Derp\n"
+                     "*-- Herp\n"
+                     "    |-- Derp\n"
+                     "    *-- Derp\n";
```
jhe-: chars not supported?

codemercenary: The unicode characters don't work properly on Windows consoles.  Had to swap it out for this stuff instead.

PR #941, opened 2016-05-11
Add static assertions to clarify incorrect use of signal

codemercenary:
Users are sometimes baffled by the cryptic error messages of `autowiring::signal` when defining handler types with non-const reference arguments.  Add static asserts to clarify these errors earlier.


wal-rus:
Awesome, thanks!

codemercenary:
Will reopen if I can get this thing working as I expect.

PR #942, opened 2016-05-18
Make toolchain filename consistent

codemercenary:
The toolchain file toolchain-arm.cmake should be renamed to toolchain-arm32.cmake in order to make it consistent with the corresponding Android toolchain filenames.


hham:
looks good.

PR #945, opened 2016-05-18
Fix memory leak in autowiring::signal

codemercenary:
Fix a memory leak caused by the `autowiring::signal` destructor not properly freeing the last listener in the list of listeners.  Also add a test to guarantee we don't miss this again, and add a corresponding test to `autowiring::once_signal` just in case.


ajohnston33:
Looks good.

PR #955, opened 2016-05-23
Add Rundown behavior to DispatchQueue

codemercenary:
This behavior is necessary to get the `CoreThread` teardown behavior to function properly.  It is also specifically mentioned in the documentation for `Abort`, so it's probably a good idea that we actually implement it.

Also use this behavior in `CoreThread::DoRunLoopCleanup`.  There is a race condition that occurs in the current implementation which can arise when a lambda is pended at about the same time as termination happens; this lambda can sometimes be leaked.


yeswalrus:
Reviewed, looks good.

PR #960, opened 2016-06-18
Eliminate unneeded default initializer

codemercenary:
This is causing VS2013 CTP to crash, and it's not necessary, remove it.


yeswalrus:
Merging optimistically.

PR #963, opened 2016-07-06
Test bad config field

RandomOutput:
Test the behavior of AutoConfig if setting an undefined field is attempted.


codemercenary:
Well, looks like it works as designed.

PR #964, opened 2016-07-06
Autoconfig/set from member

yeswalrus:
Auto config does not behave as expected when setting values directly from a field that is injected into a context.

### DIFF
--- a/src/autowiring/test/AutoConfigTest.cpp
+++ b/src/autowiring/test/AutoConfigTest.cpp
@@ -96,8 +96,11 @@ TEST_F(AutoConfigTest, ConfigFieldAssign) {
 }
 
 TEST_F(AutoConfigTest, ConfigDefault) {
+  // Default value will be whatever bUnsigned is assigned to in the inline initializer.  This is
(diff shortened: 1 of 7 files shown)


src/autowiring/test/AutoConfigTest.cpp
```diff
@@ -191,6 +202,33 @@ TEST_F(AutoConfigTest, ContextSetAfter) {
   ASSERT_EQ(mcc->b, 10442);
 }
 
+TEST_F(AutoConfigTest, ContextGetSimple) {
+  AutoCurrentContext ctxt;
+  AutoRequired<MyConfigurableClass> mcc;
+
+  mcc->b = 10442;
```
codemercenary: `MyConfigurableClass::b` is an `std::atomic`, not an observable.  The system is behaving as designed here.

codemercenary: ...except that it's crashing, I'll look into that.

yeswalrus:
Looks good to me!

PR #968, opened 2016-07-13
Add invoke feature to signal

codemercenary:
Invoke gives users a way to run an arbitrary function in the call order of a signal.  This can be useful when a small amount of setup work is necessary for a single listener, and that setup work must happen in the call order of the event being registered.

It can also be useful in running a lambda after all other signals have been registered.

One complex use case:

```C++
std::string value;
autowiring::signal<void()> onChanged;
onChanged.invoke([&] {
  // Register the handler in the signal's synchronization context.
  onChanged += [&] {
    // This gets invoked after all signal handlers for the current signal are done
    onChanged.invoke([this] {
      std::cout << "All done!" << std::endl;
    });
  };

  // Now we can also take a snapshot of the value the signal might be referencing.
  std::cout << "Value is " << value << std::endl;
});

// This might possibly be how a user would update the value without needing a mutex to do so
std::string newValue;
onChanged.invoke([&, newValue] {
  value = newValue;
});
```

### DIFF
--- a/src/autowiring/signal.h
+++ b/src/autowiring/signal.h
@@ -593,12 +593,10 @@ namespace autowiring {
     template<typename Fn>
     void invoke(Fn&& fn) {
       // For a discussion of what's happening here, see the operator() overload
-      if (!try_enter()) {
-        handoff(
+      if (!try_enter())
+        return handoff(
           new callable<Fn>{ std::forward<Fn&&>(fn) }
         );
-        return;
-      }
 
       // We've entered the asserting state succesfully.  Invoke the function and then leave the state.
       fn();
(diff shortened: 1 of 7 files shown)


src/autowiring/signal.h
```diff
@@ -348,6 +295,145 @@ namespace autowiring {
...
+    /// Attempts to enter the asserting state
+    /// </summary>
+    bool enter_asserting(void) {
+      for (;;) {
+        SignalState state = SignalState::Free;
+        if (m_state.compare_exchange_weak(state, SignalState::Asserting, std::memory_order_acquire, std::memory_order_relaxed))
+          // Entered the state successfully!
+          return true;
+
+        switch (state) {
```
hham: Should this be m_state?

yeswalrus:
Looks good, including one or two nice refactors (moving the detail stuff to it's own header)

PR #972, opened 2016-08-05
Fix NotifyWhenAutowired not working with some types

yeswalrus:
In order to work correctly the type must first be registered using auto_id_t. By adding the registration in the call to NotifyWhenAutowired, this problem is prevented at the expense of disallowing calls to NotifyWhenAutowired on forward declared types

### DIFF
--- a/src/autowiring/test/PostConstructTest.cpp
+++ b/src/autowiring/test/PostConstructTest.cpp
@@ -142,6 +142,48 @@ TEST_F(PostConstructTest, VerifyTrivialNotifyWhenAutowired) {
   ASSERT_TRUE(called) << "An autowiring notification was not invoked on an already-satisfied field as expected";
 }
 
+TEST_F(PostConstructTest, DelayedNotifyWhenAutowired) {
+  // Autorequire, and add a registration:
+  bool called = false;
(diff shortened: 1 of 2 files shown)


src/autowiring/test/PostConstructTest.cpp
```diff
@@ -156,6 +198,15 @@ TEST_F(PostConstructTest, MultiNotifyWhenAutowired) {
...
 }
 
+/*
+ Compile time failure case:
+class ForwardDeclared;
+TEST_F(PostConstructTest, TestForwardDeclare) {
+  Autowired<ForwardDeclared> decl;
+  decl.NotifyWhenAutowired([]() {std::cout << "thing"; });
+}
+*/
```
codemercenary: Delete?

yeswalrus: No - I'm not sure how we want to handle the documentation of compile time failures? I'd like to leave it in

codemercenary: Oh, I missed the part where you commented about what the failure should be.

I've thought about this, I experimented with CMake's `try_compile` feature, but decided against using that for things that aren't build environment provisioning.

PR #979, opened 2016-08-16
Fix VS2015 compiler error

codemercenary:
This is because MSVC sometimes has trouble expanding empty parameter packs into certain template substitution contexts.  Generally speaking a template specialization--such as this one--is a sure way to avoid the issue because it prevents an empty parameter pack from existing in the first place.

### DIFF
--- a/src/autotesting/AutowiringEnclosure.h
+++ b/src/autotesting/AutowiringEnclosure.h
@@ -86,6 +86,7 @@ public:
   {}
 
   const bool allowGlobalReferences;
+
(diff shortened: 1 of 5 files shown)


codemercenary:
Hold on merging this one, it still has problems in VS2015.

codemercenary:
Ok this is good to go now.

codemercenary:
Ok this is good to go now.

codemercenary:
Ok we're good to go now.

PR #980, opened 2016-08-17
Fix bug in concurrent output debugging

codemercenary:
* If running right now, we're in the RUNNING state
* If we aren't running and we should stop, we are in the STOPPED state
* If we aren't running, shouldn't stop, but we were started, we're in the STARTING state
* Otherwise we haven't been started yet and are in the WAITING state


hham:
Looks good. I am testing it on my machine.

PR #986, opened 2016-08-19
Config/float marshal

yeswalrus:
As it turns out, converting from a floating point number to a string and back is actually an exceptionally hard problem with multiple research papers published. Current state of the art for a correct, fast implementation would be Grisu3 with Dragon4 or one of it's variants as a fallback. If absolute correctness is not an requirement, Grisu2 would also be appropriate. An implementation can be found at 
https://github.com/google/double-conversion or in the RapidJSON library.

Until we have that, disable for safety.


jdonald:
AutoConfig seems to require float marshalling, so I don't think it's an option to disable it.

But as `autowiring::marshall` seems to only be used by AutoConfig, we probably don't need floating point marshalling to be perfect across denorms, quiet NaNs, or signalling NaNs either.

How about we restore the custom floating point code and bring this branch home? Then we still benefit from an integer marshalling implementation that is 30 lines smaller.

yeswalrus:
@jdonald I'll open a separate PR to use the STL functions where appropriate, but float parsing is still a serious problem and the implementation given here is actually wildly incorrect - 0.0001f is printed as 0.00001

yeswalrus:
I'm gonna cancel this PR - if we want to introduce a proper JSON or double-marshaling library we can do so later.

PR #988, opened 2016-08-28
Fix msvc14 64-bit release mode linker error

codemercenary:
We declare but never explicitly define the void `init` blocks.  Somehow this still works on every other build configuration, but not VC2015 release mode.  Fix this.


hham:
looks good.

PR #994, opened 2016-09-15
Add some convenience operators to observable

codemercenary:
These seem like they would be used frequently enough for an observable type, especially the comparison operators which are critical to the

@jdonald update: It seems we need only the hash function.

### DIFF
--- a/src/autowiring/observable.h
+++ b/src/autowiring/observable.h
@@ -1,4 +1,4 @@
-// Copyright (C) 2012-2015 Leap Motion, Inc. All rights reserved.
+// Copyright (C) 2012-2016 Leap Motion, Inc. All rights reserved.
 #pragma once
 #include "marshaller.h"
 #include "signal.h"
--- a/src/autowiring/observable.h
+++ b/src/autowiring/observable.h
@@ -72,17 +72,6 @@ public:
     onChanged();
     return *this;
   }
-
-  bool operator==(const T& rhs) { return val == rhs; }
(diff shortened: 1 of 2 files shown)


src/autowiring/observable.h
```diff
@@ -72,8 +72,18 @@ class observable {
...
+  bool operator!=(const T& rhs) { return val != rhs; }
+  bool operator<(const T& rhs) { return val < rhs; }
+  bool operator<=(const T& rhs) { return val <= rhs; }
+  bool operator>(const T& rhs) { return val > rhs; }
+  bool operator>=(const T& rhs) { return val >= rhs; }
+  auto operator*(const T& rhs) -> decltype(val * rhs) { return val * rhs; }
+  auto operator/(const T& rhs) -> decltype(val * rhs) { return val / rhs; }
+  auto operator+(const T& rhs) -> decltype(val * rhs) { return val + rhs; }
+  auto operator-(const T& rhs) -> decltype(val * rhs) { return val - rhs; }
+};
```
jdonald: These operators don't seem to be hit at all via the unit test. I mean, I can remove them and the test still compiles and runs.

I suspect it's because `operator const T&(void) const { return val; }` already does everything we need.

Still investigating.

src/autowiring/observable.h
```diff
@@ -100,3 +99,16 @@ struct marshaller<observable<T>> :
 };
 
 }
+
+namespace std {
```
yeswalrus: Nice addition, this will be handy

PR #995, opened 2016-09-16
Add a way to determine whether a signal has listeners

codemercenary:
Some signals are actually quite expensive to assert.  It would be helpful to be able to query a signal to determine whether it has listeners before doing work to try to invoke it.

### DIFF
--- a/src/autowiring/test/AutoSignalTest.cpp
+++ b/src/autowiring/test/AutoSignalTest.cpp
@@ -819,3 +819,10 @@ TEST_F(AutoSignalTest, MoveInInvoke) {
   ASSERT_NE(nullptr, vRecovered);
   ASSERT_EQ(404, *vRecovered) << "Recovered unique pointer was not the expected value";
 }
+
+TEST_F(AutoSignalTest, ListenerCheck) {
+  autowiring::signal<void()> s;
+  ASSERT_FALSE(s) << "Signal did not correctly report it had no listeners";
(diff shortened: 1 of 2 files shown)


src/autowiring/test/AutoSignalTest.cpp
```diff
@@ -819,3 +819,10 @@ TEST_F(AutoSignalTest, MoveInInvoke) {
   ASSERT_NE(nullptr, vRecovered);
   ASSERT_EQ(404, *vRecovered) << "Recovered unique pointer was not the expected value";
 }
+
+TEST_F(AutoSignalTest, ListenerCheck) {
+  autowiring::signal<void()> s;
+  ASSERT_FALSE(s) << "Signal did not correctly report it had no listeners";
+  s += [] {};
+  ASSERT_TRUE(s) << "Signal stated that it had no listeners when it should have had at least one";
```
yuwilbur: Perhaps you can add one more additional use case such that:
```
auto listener = [] {};
auto registration = s += listener;
ASSERT_TRUE(s) << "Signal stated that it had no listeners when it should have had at least one";
s -= resgistration;
ASSERT_FALSE(s) << "Signal stated that it has listeners after removing all listeners";
```

codemercenary: Good idea.

yuwilbur:
Looks good!

PR #996, opened 2016-09-18
Fix null dereference in PrintRunnables

codemercenary:
This routine prints context member names directly.  If the name happens to be nullptr, this causes an exception.  Add a test and fix this mistake.

### DIFF
--- a/src/autowiring/AutowiringDebug.cpp
+++ b/src/autowiring/AutowiringDebug.cpp
@@ -188,8 +188,12 @@ void autowiring::dbg::PrintRunnables(std::ostream& os, CoreContext& ctxt) {
 
         // Type information, in human-readable form
         os << ' ' << autowiring::demangle(typeid(*runnable)) << ' ';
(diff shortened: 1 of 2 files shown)


src/autowiring/AutowiringDebug.cpp
```diff
@@ -188,8 +188,12 @@ void autowiring::dbg::PrintRunnables(std::ostream& os, CoreContext& ctxt) {
 
         // Type information, in human-readable form
         os << ' ' << autowiring::demangle(typeid(*runnable)) << ' ';
-        if (ContextMember* pMember = dynamic_cast<ContextMember*>(runnable))
-          os << " (" << pMember->GetName() << ')';
+        if (ContextMember* pMember = dynamic_cast<ContextMember*>(runnable)) {
+          if (const char* name = pMember->GetName())
```
jdonald: Cool, didn't know variables can now be declared inside the `if ()` block like so.

jdonald:
Unit test passes, looks good.

PR #997, opened 2016-09-18
Fix incorrect WaitForStateUpdate behavior

codemercenary:
This routine must throw an exception if the thread is terminated during a wait.  Ensure that `BasicThread::OnStop` signals the state mutex when it's called to cause any waiters to wake up.


hham:
Looks good.

PR #998, opened 2016-09-18
Add self-descriptive version number

codemercenary:
Sometimes workarounds are required downstream.  Provide definitions of the current version number so that these workarounds can be identified with a `static_assert`, and removed when the version number is updated.


jdonald:
I managed to clean this up. There were two major bugs as of this morning.
- Adding `"$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>"` caused an obscure build error that only manifested with in-source builds. We had been exporting a header called `signal.h` which naturally conflicts with the standard UNIX header of the same name.
- The InstallHeaders.cmake changes were never tested, and broke `make install` such that it would manifest when building the examples child project. I've hotfixed here and put out a corresponding PR in the `standard` repo.

@yuwilbur and @jhe- you're among the assignees. Did either of you request this feature? Come to think of it, it's pretty handy given we sometimes leave autowiring bug workarounds in platform longer than we need to.

src/autowiring/test/CMakeLists.txt
```diff
@@ -94,6 +94,7 @@ set(AutowiringFixture_SRCS
 
 add_library(AutowiringFixture ${AutowiringFixture_SRCS})
 add_pch(AutowiringFixture stdafx.h stdafx.cpp)
+target_link_libraries(AutowiringFixture Autowiring)
```
yeswalrus: Good catch

jhe-:
Looks good to me. I didn't request the feature but I see its utility.

yeswalrus:
Fixes a bunch of stuff beyond what it says it does, and all of it is useful. Nice work James, thank you!

jdonald:
It turns out a couple headers in platform are explicitly including `<autowiring/signal.h>`.

That header now (now called `<autowiring/auto_signal.h>` is implicitly included by CoreContext,h and other headers though, so I'll remove those lines from platform.

PR #999, opened 2016-09-21
Remove legacy support for Xcode + libstdc++

jdonald:
We soon need to update autowiring to support Xcode 8. In the meantime, remove legacy unsupported configurations such as the libstdc++ build.

Removing the `_LIBCPP_VERSION` leads to implying `STL11_ALLOWED` for all of MSVC/clang/gcc and simplifies much of this header.

I tried to leave support for older versions of **gcc** intact, although I suspect we're not testing those paths either.


src/autowiring/C++11/cpp11.h
```diff
@@ -2,25 +2,17 @@
...
-    #define STL11_ALLOWED 1
-  #else // clang + libstdc++
-    #define STL11_ALLOWED 0
-  #endif
-#else // gcc or MSVC
-  #define STL11_ALLOWED 1
-#endif
+// various compiler targets.
+
+#define STL11_ALLOWED 1
```
hham: STL11_ALLOWED seems unnecessary.

hham:
It looks good except for unnecessary STL11_ALLOWED define.

src/autowiring/C++11/cpp11.h
jdonald: Oh that was done out of caution given existing projects could rely on it. In platform, I removed one use of it in FileSystemUtil.cpp only recently.

Nonetheless, we probably shouldn't have users relying on it. I just checked and neither leapipc nor LeapHTTP use it, so I've gone ahead and removed here.

hham:
Good to move on.

PR #1000, opened 2016-09-21
Update copyright notices to 2016

jdonald:
My other pull request was failing on `scripts/copyright_check.sh` because I was diligent in updating the header comments...

We need to fix the script, which then requires fixing the affected (300+ !) files. Once this goes in, unfortunately all active pull requests will have to be stabilized in order to pass Travis.


hham:
Looks good.

PR #1001, opened 2016-09-28
Update CMake standard

jdonald:
So apparently leaphttp and leapipc depend on autowiring now. And Autowiring 1.0.3 had the bug of losing fat binary support.

So now we need to update to the latest leapmotion/standard for Autowiring 1.0.4 in order to get the other repos fixed (since they wouldn't be able to link with a non-fat binary autowiring even when they update to the latest leapmotion/standard).

- [x] Updated `leapmotion/standard/master`


hham:
AutowiringTest works fine.

PR #1003, opened 2016-09-29
Config/stl marshal

yeswalrus:
Eliminate custom parsing/marshaling code and use the STL where appropriate.

### DIFF
--- a/src/autowiring/marshaller.h
+++ b/src/autowiring/marshaller.h
@@ -155,7 +155,7 @@ namespace autowiring {
   template<typename T>
   struct float_converter;
 
-#ifdef MSVC
+#if defined(_MSC_VER) && _MSC_VER <= 1900 //This is reported resolved in VC++15
 #define AW_SNPRINTF sprintf_s
 #else
 #define AW_SNPRINTF snprintf
--- a/src/autowiring/marshaller.h
+++ b/src/autowiring/marshaller.h
@@ -9,6 +9,7 @@
 #include <string>
 #include <string.h>
 #include <stdlib.h>
+#include <stdio.h>
 #include TYPE_TRAITS_HEADER
 
 namespace autowiring {
@@ -156,17 +157,17 @@ namespace autowiring {
 
   template<>
   struct float_converter<float> {
-    static float convert(const char* szValue) { return std::strtof(szValue, nullptr); }
+    static float convert(const char* szValue) { return strtof(szValue, nullptr); }
   };
 
   template<>
   struct float_converter<double> {
-    static double convert(const char* szValue) { return std::strtod(szValue, nullptr); }
+    static double convert(const char* szValue) { return strtod(szValue, nullptr); }
   };
 
   template<>
   struct float_converter<long double> {
-    static long double convert(const char* szValue) { return std::strtold(szValue, nullptr); }
+    static long double convert(const char* szValue) { return strtold(szValue, nullptr); }
   };
 
   template<typename T>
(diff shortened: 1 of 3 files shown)


jdonald:
Build errors with gcc and clang.

yeswalrus:
@jdonald Resolved and improved

src/autowiring/test/MarshallerTest.cpp
```diff
@@ -6,12 +6,54 @@ class MarshallerTest:
...
+  float outX = 20;
+  f.unmarshal(&outX, valX.c_str());
+  ASSERT_EQ(outX, 0) << "Failed to properly round trip 0";
+
+  float y = 0.0001f;
+  std::string valY = f.marshal(&y);
+
+  //This value is actually notoriously difficult to create a minimal string
+  //representation for.
+  //ASSERT_STREQ("0.0001", valY.c_str()) << "Failed to marshal a decimal value";
```
jdonald: dead code?

src/autowiring/marshaller.h
```diff
@@ -171,104 +153,66 @@ namespace autowiring {
   };
 
   template<typename T>
+  struct float_converter;
+
+#ifdef MSVC
```
jdonald: I think the macro you are looking for is `_MSC_VER`. `MSVC` isn't normally defined.

If it's working despite this bug, maybe all you need is `snprintf()` for all platforms. But on the other hand, Travis doesn't test with Visual Studio.

yeswalrus: Unfortunately, snprintf is undefined in MSVC2013 and 2015, only having been added in VC "15"

jdonald: In that case you may want to make the condition to be `defined(_MSC_VER) && _MSC_VER <= 1900`

src/autowiring/test/MarshallerTest.cpp
yeswalrus: Removed

src/autowiring/marshaller.h
yeswalrus: Done.

src/autowiring/test/AutoConfigTest.cpp
```diff
@@ -5,6 +5,7 @@
 #include <autowiring/ConfigRegistry.h>
 #include <autowiring/observable.h>
 #include <cstring>
+#include "stdlib.h"
```
yeswalrus: Fixed

src/autowiring/test/AutoConfigTest.cpp
```diff
@@ -166,15 +168,17 @@ TEST_F(AutoConfigTest, Double) {
 
   c.e = 10929.4423;
   strVal = autowiring::ConfigGet("e", c);
-  ASSERT_STREQ("10929.4423", strVal.c_str());
+  double outVal = strtod(strVal.c_str(),nullptr);
```
yeswalrus: Fixed

jdonald:
Thanks for fixing all that.

I was about to merge, then when I browsed the very first commit I saw it said `std::to_string`, and which reminded me that the NDK doesn't support this for 32-bit targets (which we must continue to support for Baofeng).

```
/home/jdonald/autowiring/src/./autowiring/marshaller.h:140:14: error: 'to_string' is not a member of 'std'
       return std::to_string(val);
```

yeswalrus:
:/ ok, I'll handle it

yeswalrus:
Fixed - is it possible for us to automate checks for disallowed STL functions?

jdonald:
That last error is probably a random failure in AutoConfig_SliderTest.ConcurrentModification, but this new implementation seems unnecessarily complex.

The `stringstream` 2-line approach seemed fine. It's actually used in many other places throughout Autowiring (as Jason did intentionally to keep things compatible with Android). Why not stick with that instead of the new template? The custom implementation is harder to verify, and opens up more questions which I'll comment on inline.

src/autowiring/marshaller.h
```diff
@@ -128,34 +129,40 @@ namespace autowiring {
...
+    auto len = AW_SNPRINTF(buffer, sizeof(buffer), fmt, value);
+    return std::string(buffer, len);
+  }
+#undef AW_SNPRINTF
+
+  static inline std::string int_to_string(char c) { return var_to_string<char, 8>(c, "%hhd"); }
+  static inline std::string int_to_string(short c) { return var_to_string<short, 8>(c, "%hd"); }
+  static inline std::string int_to_string(int c) { return var_to_string<int, 16>(c, "%d"); }
+  static inline std::string int_to_string(long c) { return var_to_string<long, 32>(c, "%ld"); }
+  static inline std::string int_to_string(long long c) { return var_to_string<long long, 64>(c, "%lld"); }
```
jdonald: 64? This template arg appears to be the size of the target buffer. For the minimum value of a `long long`, around negative 8 quintillion, I calculate that to be around 22 including the negative sign and null terminator, much less than 64.

src/autowiring/marshaller.h
```diff
@@ -128,34 +129,40 @@ namespace autowiring {
...
+    char buffer[bufferLen];
+    auto len = AW_SNPRINTF(buffer, sizeof(buffer), fmt, value);
+    return std::string(buffer, len);
+  }
+#undef AW_SNPRINTF
+
+  static inline std::string int_to_string(char c) { return var_to_string<char, 8>(c, "%hhd"); }
+  static inline std::string int_to_string(short c) { return var_to_string<short, 8>(c, "%hd"); }
+  static inline std::string int_to_string(int c) { return var_to_string<int, 16>(c, "%d"); }
+  static inline std::string int_to_string(long c) { return var_to_string<long, 32>(c, "%ld"); }
```
jdonald: This should use the same buffer size as a `long long`, as it's equivalent on gcc/clang.

jdonald:
> is it possible for us to automate checks for disallowed STL functions or at least keep a list of them somewhere and why they're disallowed?

You could write and maintain a lint script (autowiring already has a few under `scripts`), but `std::to_string` is only part of the larger problem that we aren't running an Android build on Travis.

This wasn't possible when we investigated in September 2015, but it appears Android support for Travis is now in beta: https://docs.travis-ci.com/user/languages/android/

yeswalrus:
I went with the new template code for parity with the other parts where we were doing printing with  sprintf, and to avoid pulling stringstream into a core header since it's notoriously heavyweight.

jdonald:
Good call on trying to reduce header overhead. This gets included by `<autowiring/config.h>`, but at least not other Autowiring core headers, right? Even so, one mitigation strategy would be to do `#if __ANDROID__` for stringstream and use `std::to_string()` in all other cases.

(I was incorrect earlier in saying this was specific to 32-bit Android. I reproduced the error message with the aarch64 ndk.)

Understood, you tried sprintf with integers because that's what worked for floats. I get that it's not possible to use `std::to_string()` with arbitrary precision on floats. Though to me it seems preferable to be inconsistent if that keeps each implementation simple.

yeswalrus:
Yeah, I'd prefer brevity if it's possible. I'll go for that #if __ANDROID solution. Also, I've been unable to reproduce that concurrency problem on mac, even with gtest_repeat=1000 and adding some waits in a couple places

jdonald:
Verified to build on Android.

After testing `AutoConfig_SliderTest` under Release, Debug, valgrind, etc, for now we'll have to chalk it up to an Xcode 6 optimization bug.

jdonald:
I made the mistake of only testing Android 64-bit. It turns out we don't have `strtolod()` with the 32-bit NDK and will have to fix this.

yeswalrus:
OK. Strstream can be made to do the job

PR #1006, opened 2016-10-12
Fix Android 32-bit build

jdonald:
Here's the fix that @yeswalrus recommended, using stringstream.

I've clarified the preprocessor conditions so we have one less issue when eventually upgrading from gnustl to libc++.

Travis doesn't build Android yet, so I verified both 32-bit and 64-bit builds locally.

- [x] Fix Travis clang build first
- [x] Check MSVC
- [x] Fixes the issue introduced in PR #1003

### DIFF
--- a/src/autowiring/marshaller.h
+++ b/src/autowiring/marshaller.h
@@ -12,7 +12,7 @@
 #include <stdio.h>
 #include TYPE_TRAITS_HEADER
 
-#if __ANDROID__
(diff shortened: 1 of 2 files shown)


yeswalrus:
@hham I checked MSVC, as soon as you verify this fixes the android 32 bit builds it should be good to go.

hham:
Oh good. I was building on my Windows laptop.

jdonald:
Thanks guys!

PR #1008, opened 2016-10-12
Restore <autowiring/signal.h>

jdonald:
While `<signal.h>` can be confusing as it shares its name with a standard UNIX header, it's explicit enough when customers `#include <autowiring/signal.h>`

We have good reason for calling it `<autowiring/signal.h>` as outlined in #851 when this was first added. Furthermore, I don't want to make a breaking change forcing us to autowiring-1.1.

The problem we had to work around in #998 is properly dealt with by not adding all the dangerous include paths.

- [x] Review #1009 first

### DIFF
--- a/src/autowiring/CMakeLists.txt
+++ b/src/autowiring/CMakeLists.txt
@@ -245,7 +245,7 @@ target_include_directories(
   "${PROJECT_SOURCE_DIR}/contrib/autoboost"
   PUBLIC
   "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>"
-  "$<BUILD_INTERFACE:${PROJECT_BUILD_DIR}/src>"
+  "$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/src/autowiring>"
   INTERFACE
   "$<INSTALL_INTERFACE:include>"
 )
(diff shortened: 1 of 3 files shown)


src/autowiring/CMakeLists.txt
```diff
@@ -244,7 +245,7 @@ target_include_directories(
   "${PROJECT_SOURCE_DIR}/contrib/autoboost"
   PUBLIC
   "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>"
-  "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>"
+  "$<BUILD_INTERFACE:${PROJECT_BUILD_DIR}/src>"
```
hham: This causes a build error, "AutowiringVersion.h: No such file or directory."

jdonald: It's passing on Travis though . I have tested both in-source and out-of-source builds.

What are the steps to reproduce the error you're seeing?

hham: I made autowiring-build directory and ccmake ../autowiring inside the directory. PROJECT_BUILD_DIR is null in my Mac and Linux box.

jdonald: Fixed. Hyoung and I figured it out it was working on Travis only because that tests in-source builds.

hham:
Build fails. It seems to be from travis setting.

jdonald:
I've pushed a new fix that seems to work for Travis's in-source build. Please try it out on your out-of-source build environment.

hham:
Builds fine in my local machine and travis.

PR #1009, opened 2016-10-13
Update standard

jdonald:
We had several fixes to `standard` today, but the main one we need is fixing the clang build by properly specifying the OS X SDK.

This would fix our Travis builds to unblock all the other open autowiring pull requests.


hham:
Looks good.

PR #1011, opened 2017-05-16
Actually use pch.

yeswalrus:
Previous pch code did not work correctly


jdonald:
Awesome, verified.

image

and in the command line box, `/Yu"stdafx.h"` appears in the beginning rather than the middle.

compared to the baseline which says Not using precompiled heeaders.

PR #1012, opened 2017-05-17
Fix vs2015 type deduction error

yeswalrus:
Changes to the way msvc handles template deduction caused a break. I believe this is the correct fix.

### DIFF
--- a/src/autowiring/test/AutoConstructTest.cpp
+++ b/src/autowiring/test/AutoConstructTest.cpp
@@ -102,3 +102,47 @@ TEST_F(AutoConstructTest, FactoryNewPrivateCtor) {
   ASSERT_NE(nullptr, mpcc.get()) << "Null not expected as a return type from a factory new construction";
   ASSERT_EQ(1002, mpcc->ival) << "Correct ctor was not invoked on a type with a private ctor";
 }
+
+
+namespace {
+  class MyPrivateCtorStringClass :
+    public CoreObject
+  {
+    MyPrivateCtorStringClass(void) :
+      istr("default_string")
+    {}
+    MyPrivateCtorStringClass(const char*) :
+      istr(istr)
+    {}
+
+  public:
+    const char* istr;
+
+    static MyPrivateCtorStringClass* New(const char* str) {
+      return new MyPrivateCtorStringClass{ str };
+    }
+  };
+}
+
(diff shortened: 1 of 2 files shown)


src/autowiring/test/AutoConstructTest.cpp
```diff
@@ -102,3 +102,46 @@ TEST_F(AutoConstructTest, FactoryNewPrivateCtor) {
...
+
+namespace {
+  class MyPrivateCtorStringClass :
+    public CoreObject
+  {
+    MyPrivateCtorStringClass(void) :
+      istr("default_string")
+    {}
+    MyPrivateCtorStringClass(const char*) :
+      istr(istr)
```
gittyupagain: The member variable`istr` ends up being uninitialized here, as you are ignoring the parameter passed to the ctor.

src/autowiring/test/AutoConstructTest.cpp
```diff
@@ -102,3 +102,46 @@ TEST_F(AutoConstructTest, FactoryNewPrivateCtor) {
...
+  !std::is_constructible<MyPrivateCtorStringClass>::value,
+  "Type reported as being constructable when it was not"
+  );
+static_assert(
+  !autowiring::has_simple_constructor<MyPrivateCtorStringClass>::value,
+  "Simple constructor detected when said constructor should have been private"
+  );
+
+TEST_F(AutoConstructTest, FactoryNewPrivateStringCtor) {
+  AutoConstruct<MyPrivateCtorStringClass> mpcc{ "a new string" };
```
gittyupagain: The problem with Visual Studio 2015 appears to be with converting from `const char(&)[13]` to `const char*`. There may be a problem with Visual Studio 2015, but that isn't a reason to remove the perfect forwarding. This is fixed in Visual Studio 2017, BTW. If you use cast the `"a new string"` to a `const char*`, the perfect forwarding works in Visual Studio 2015 as well.

yeswalrus:
Gotcha. Looks like this isn't a problem in autowiring after all..

jdonald:
> If you use cast the "a new string" to a const char*, the perfect forwarding works in Visual Studio 2015 as well.

Sweet, that gives us another workaround if we decide to retain compatibility with Visual Studio 2015.

PR #1013, opened 2017-05-18
Fix has_static_new

yeswalrus:
MSVC2015, alone among MSVC versions and for reasons that are beyond me chokes on the function name detection portion of the old has_static_new's SFINAE.
This overhauls it to be far more concise, and compile correctly.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+
+[**]
+indent_style = space
+indent_size = 2
+insert_final_newline = true
--- a/.gitignore
+++ b/.gitignore
@@ -42,8 +42,6 @@ Autowiring.nuspec
 autowiring.kdev4
 *.opendb
 *.orig
-autowiring.VC.db
+*.VC.db*
 .vscode
 AutowiringVersion.h
-
-*.VC.db*
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,5 @@ autowiring.kdev4
 autowiring.VC.db
 .vscode
 AutowiringVersion.h
+
+*.VC.db*
--- a/src/autowiring/has_static_new.h
+++ b/src/autowiring/has_static_new.h
@@ -32,7 +32,7 @@ template<typename T, typename... Args>
 struct has_static_new
 {
   template<class U>
-  static std::true_type select(decltype(U::New((Args&&)*(Args*)nullptr...))*);
+  static std::true_type select(decltype(U::New(std::declval<Args>()...)));
 
   template<class U>
   static std::false_type select(...);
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -45,3 +45,5 @@ autowiring.kdev4
 autowiring.VC.db
 .vscode
 AutowiringVersion.h
+
+*.VC.db*
```
jdonald: `Autowiring.VC.db is already listed above though?

Or if the intent here is the trailing asterisk, are there circumstances when it makes a db1, db2 file?

src/autowiring/test/AutoConstructTest.cpp
```diff
@@ -102,3 +107,65 @@ TEST_F(AutoConstructTest, FactoryNewPrivateCtor) {
...
+  "Found a bad static new"
+  );
+static_assert(
+  autowiring::has_static_new<MultiStaticNew, int>::value,
+  "Failed to find a good static new override"
+  );
+
+TEST_F(AutoConstructTest, MultiStaticNewTest) {
+  AutoConstruct<MultiStaticNew> badstatic{ 1 };
+}
```
jdonald: missing newline at EOF

jdonald: This var is named `badstatic`. However, it looks like the good one?

src/autowiring/has_static_new.h
```diff
@@ -5,51 +5,18 @@
 
 namespace autowiring {
 
-/// <summary>
-/// Utility helper structure for types which have a factory New routine
-/// </summary>
-/// <remarks>
-/// A factory New routine is malformed when the return type is not implicitly castable to type T
-/// </remarks>
```
jdonald: Should keep the comments? I found the `<remarks>` line useful initially.

jdonald:
This implementation is effectively `-38 +10`. I've tested platform on OS X and MSVC2013 and it probably works everywhere. Great work, @yeswalrus !

.gitignore
yeswalrus: Yeah, 2017 adds several extensions.

jdonald: I see now. I have `Browse.VC.db` and `Solution.VC.db{,-shm,-wal}`.

Should remove the now-redundant `autowiring.VC.db` above then?

.editorconfig
```diff
@@ -0,0 +1,6 @@
+root = true
+
+[**]
+indent_style = space
+indent_size = 2
+insert_final_newline = true
```
jdonald: Ironic that this file is missing its final newline!

yeswalrus: Sadly, VS2017 has default support for editor config but ignores the 2 most useful options: insert_final_newline and trim_trailing_whitespace

.gitignore
yeswalrus: yep

PR #1014, opened 2017-05-23
Fix gcc issue with constructors taking references to abstract objects

yeswalrus:
Uncovered when linking with platform on arm-gcc5.


src/autowiring/test/AutoConstructTest.cpp
```diff
@@ -169,3 +169,4 @@ static_assert(
 TEST_F(AutoConstructTest, MultiStaticNewTest) {
   AutoConstruct<MultiStaticNew> multistatic{ 1 };
 }
+
```
jdonald: Stray edit?

jdonald:
Confirmed AutoConstruct (which I has_static_new touches) still works on platform and passes some tests (at least on Mac and MSVC2013) when autowiring has this change.

PR #1015, opened 2017-05-24
Remove unused config file

yeswalrus:
This file appears to be a leftover remnant of some autogenerating code from long ago that was never removed


jdonald:
Confirmed we aren't referencing `AutowiringConfig.h` in platform, leapserial, LeapResource, leaphttp, os-controls (other than an embedded subrepo).

PR #1016, opened 2017-05-24
Upd standard

yeswalrus:
Pull changes from standard


jdonald:
Looks the same as what I tested locally on macOS the other day 👍

PR #1017, opened 2017-05-25
Updated autoboost to 1.64

yeswalrus:
As the title says - 1.64 is the first release fully supporting MSVC2017, in addition to having a number of bugfixes


jdonald:
I recall we shouldn't require autoboost now that we've deprecated Android 32-bit. Or is there some code that still references it on other platforms?

yeswalrus:
Interesting. I've been running into all kinds of obsure edge cases when trying to run with the upgraded version, so I think I'll close this pr. We should look into deprecating it when we finish upgrading our supported compilers

jdonald:
Ok so it turns out AutoNet still depends on autoboost.

We get a ton of warnings  with "Unknown compiler version" with Visual Studio 2017. We could fix that with one line more in visualc.hpp after this, but it would make sense to upgrade before doing so.

This PR is a whopping 200,000 lines. Maybe we should see if AutoNet still works on Windows and decide what to do.

jdonald:
Let's hold off on this for the time being as the implications aren't known and we want v1.0.5 sooner rather than later.

jdonald:
Closing this for now. We will likely have to revisit at some point when we start using toolchains that no longer work with boost 1.63.0.

The autoboost requirement will be going away for Android at some point, but unfortunately we don't have a plan to remove the Boost dependency from the Autonet visualizer.

PR #1018, opened 2017-06-01
Upd standard

yeswalrus:
Fixes a pair of bugs


jdonald:
I tested this last time and it seemed to pass, but obviously I missed something. Would be good for @gittyupagain to test.

PR #1019, opened 2017-06-09
Fix Visual Studio 2017 32 bit builds

yeswalrus:
workaround alignment issues with gtest


jdonald:
Does this have to be a C-style cast? `static_cast<>` doesn't work?

jdonald:
What `atomicobj.load()` work?

PR #1021, opened 2017-07-13
Avoid undefined behavior of defined()

jdonald:
When upgrading to clang for Android, we get this on virtually every source file:
```
In file included from /home/tombuilder/external-libraries/src/autowiring/src/autowiring/test/stdafx.h:7:
/home/tombuilder/external-libraries/src/autowiring/src/./autowiring/C++11/cpp11.h:11:5: warning: macro expansion producing
      'defined' has undefined behavior [-Wexpansion-to-defined]
#if IS_CLANG && !CLANG_CHECK(3, 6)
```

We should probably go ahead and fix this as clang is correct to point out that it's not part of the C preprocessor standard: https://gcc.gnu.org/onlinedocs/cpp/Defined.html

### DIFF
--- a/src/autowiring/C++11/cpp11.h
+++ b/src/autowiring/C++11/cpp11.h
@@ -4,7 +4,11 @@
 // The reason this header exists is due to the asymmetric availability of C++11 on our
 // various compiler targets.
 
-#define IS_CLANG defined(__clang_major__)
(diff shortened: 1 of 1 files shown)


hham:
Good to know the preprocessor standard.

PR #1025, opened 2017-10-26
Update to latest json11 with patches

gittyupagain:
To resolve issues with Clang and the version of the json11 library included in the contrib directory, update to https://github.com/dropbox/json11/commit/ec4e45219af1d7cde3d58b49ed762376fccf1ace, plus a Windows-specific patch that was applied to the previous version.


jdonald:
Confirmed AutoNet still works as far as I can tell.

jdonald:
I just discovered the hard way: this doesn't compile on Visual Studio 2013.

PR #1026, opened 2017-10-26
Avoid pthread_setname_np()

jdonald:
We still support the thread names on x86_64 Linux and Android, but it isn't available on RV1108.


hham:
It is mysterious that this part has been missing by now.

PR #1027, opened 2017-10-27
Fix CoreJob race condition

gittyupagain:
It was possible for current event handler to be deleted twice, and also for it to inadvertently be started multiple times, causing a few user functions to be run in different threads.

One of the main issues was that, although `CoreJob`'s `OnPended` function expected a lock, two different locks were being passed from different threads to the function, nullifying exclusivity. Also, the `wait` functions could be called from different threads, potentially causing a double free.

A test (`CoreJobTest.PendFromMultipleThreads`) has been added, but it doesn't fail every time without this fix. To see the failure case, the test should be repeated multiple times (using `--gtest_repeat`, specifying the number of runs, and `--gtest_throw_on_failure`, to stop on failure).

### DIFF
--- a/src/autowiring/CoreJob.cpp
+++ b/src/autowiring/CoreJob.cpp
@@ -99,7 +99,7 @@ bool CoreJob::OnStart(void) {
 
   m_running = true;
 
-  std::unique_lock<std::mutex> lk;
+  std::unique_lock<std::mutex> lk(m_dispatchLock);
   if(m_pHead)
     // Simulate a pending event, because we need to set up our async:
     OnPended(std::move(lk));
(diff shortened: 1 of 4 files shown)


src/autowiring/CoreJob.cpp
```diff
@@ -52,8 +52,10 @@ void CoreJob::OnPended(std::unique_lock<std::mutex>&& lk){
     // Need to ask the thread pool to handle our events again:
     m_curEventInTeardown = false;
 
-    if (m_curEvent)
-      delete static_cast<std::future<void>*>(m_curEvent);
+    std::future<void>* future = static_cast<std::future<void>*>(std::atomic_exchange<void*>(&m_curEvent, nullptr));
```
jdonald: fancy

PR #1029, opened 2017-10-30
Update copyright year to 2017

jdonald:
If we do a v1.0.5 release, we should probably update the copyright year before 2017 is over.


hham:
Looks good.

PR #1030, opened 2017-10-31
Fix parallel class

gittyupagain:
Threads were leaking due to `DispatchQueue::WaitForEvent` waiting forever for each thread. The fix was to `Rundown` the dispatch queue, causing `WaitForEvent` to abort both existing and future calls.

### DIFF
--- a/src/autowiring/Parallel.cpp
+++ b/src/autowiring/Parallel.cpp
@@ -30,7 +30,7 @@ parallel::parallel(CoreContext& ctxt, size_t concurrency):
   while (concurrency--)
     std::thread(
       [block] {
-        while(block->owned)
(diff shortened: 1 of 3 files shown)


src/autowiring/Parallel.cpp
```diff
@@ -56,7 +51,10 @@ parallel::~parallel(void) {
 }
 
 void parallel::stop(void) {
-  std::lock_guard<std::mutex> lk(m_block->m_lock);
+  auto block = m_block;
+  if (!block)
+    return;
+  std::lock_guard<std::mutex> lk(block->m_lock);
```
jdonald: 👍

PR #1033, opened 2017-11-06
Miscellaneous  minor fixes

gittyupagain:
Fix the Parallel code race condition. Workaround warnings with Clang. Clean up tests.


jdonald:
This is a dramatic improvement to autowiring stability, resolving longstanding issues with dispatch queues and AutoFilter packets.

PR #1038, opened 2018-01-04
Update copyright year

jdonald:
Might as well do this early in January rather than later! And avoid getting in the way of the next release.


hham:
Looks good.

PR #1039, opened 2018-01-05
Use std::tuple where available

yeswalrus:
Replace custom tuple implementations with std::tuple & related types.  Retains the old code as a shim for older versions of visual studio

- [ ] Testing on Linux
- [ ] Testing on Mac
- [ ] Testing on VS2015


yeswalrus:
Aborting - too much hassle creating compatibility shims for partial implementations in various gcc versions. We can pursue this again when we don't need to support pre-c++14 compilers anymore

PR #1040, opened 2018-01-06
Fix a compiler error on VS2017

yeswalrus:
/permissive- would work as well

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,9 +23,8 @@ endif()
 
 # Enable conformance mode for newer versions of MSVC
 if(MSVC_VERSION GREATER 1910)
-  string(APPEND CMAKE_CXX_FLAGS " /d1parsePackExpressions-")
+  string(APPEND CMAKE_CXX_FLAGS " /permissive-")
 endif()
-
 
 add_definitions(-DGTEST_HAS_TR1_TUPLE=0)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,12 @@ if(NOT MSVC OR MSVC_VERSION LESS 1900)
   )
 endif()
 
+# Enable conformance mode for newer versions of MSVC
+if(MSVC_VERSION GREATER 1910)
+  string(APPEND CMAKE_CXX_FLAGS " /d1parsePackExpressions-")
+endif()
+
+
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -21,6 +21,12 @@ if(NOT MSVC OR MSVC_VERSION LESS 1900)
   )
 endif()
 
+# Enable conformance mode for newer versions of MSVC
```
jdonald: `/d1parsePackExpressions-` does not enable conformance mode (under Project Settings -> C/C++ -> Language). It only affects the new code for fold expressions.

We should either update the comment or the option. It might make sense to use `/permissive-` if we can. We'd like to switch platform over to conformance mode soon and autowiring could be a good first step.

CMakeLists.txt
```diff
@@ -21,6 +21,12 @@ if(NOT MSVC OR MSVC_VERSION LESS 1900)
...
+if(MSVC_VERSION GREATER 1910)
+  string(APPEND CMAKE_CXX_FLAGS " /d1parsePackExpressions-")
+endif()
+
+
```
jdonald: extra newline here?

src/autowiring/C++11/filesystem.h
```diff
@@ -2,9 +2,13 @@
...
-#if defined(_MSC_VER) && _MSC_VER >= 1900 //weirdly, only MSVC 2015 supports this so far.
-#include <filesystem>
-namespace awfsnamespace = std::tr2::sys;
+#if defined(_MSC_VER)
+  #include <filesystem>
+  #if _MSC_VER >= 1910
+    namespace awfsnamespace = std::experimental::filesystem;
+  #elif _MSC_VER >= 1900
+    namespace awfsnamespace = std::tr2::sys;
+  #endif
```
hanzou-x: I think this requires an #else clause for earlier versions of Visual Studio to use autoboost. Unless we're intentionally deprecating them in which case there should be an #error check.

jdonald: I think this requires an #else clause for earlier versions of Visual Studio to use autoboost. Unless we're intentionally deprecating them in which case there should be an #error check.

jdonald:
This looks good, but something strange is going on with Travis. Investigating.

PR #1041, opened 2018-01-25
Fix a warning in clang 5

yeswalrus:
-Wc++1z-compat-mangling doesn't seem to exist in the clang5 that ships with NDK 16b


jdonald:
Looks good to me and works on Mac.

PR #1043, opened 2018-01-26
Alignment make share fix

bstastnyleapmotion:
@jdonald

### DIFF
--- a/src/autowiring/AutoPacket.h
+++ b/src/autowiring/AutoPacket.h
@@ -531,6 +531,7 @@ public:
   template<class T, typename... Args>
   const T& Emplace(Args&&... args) {
     static_assert(!std::is_pointer<T>::value, "Can't decorate using a pointer type.");
+
     // Create a copy of the input, put the copy in a shared pointer
     auto ptr = std::shared_ptr<T>(new T(std::forward<Args&&>(args)...));
     Decorate(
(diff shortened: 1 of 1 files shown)


jhe-:
Nice fix, @bstastnyleapmotion. Obviously not a showstopper, but does using new here instead of make_shared cause an extra copy to happen? I recall we had a similar tradeoff to make between push_back and emplace_back.

bstastnyleapmotion:
Yeah it has an extra copy but the compiler should optimize it to minimize the cost. (hopefully)

jdonald:
Bummer given Autowiring doesn't have the best record on getting these things right with performance. Let's have @gittyupagain advise on this as well.

jhe-:
@gittyupagain profiled and didn't find any noticeable increase in running time. Correct me if I'm wrong but it seems that changing to brace initialization would only have stylistic effect.

As for why we didn't encounter this before, I believe this was just due to not _directly_ using any Eigen members as autofilter types before:

```
Jimmy He [16:52]
It's possible these tests do fail sporadically in 32-bit if they have Eigen members.

James Donald [17:01]
It just doesn't match what I've seen in all the longstanding test failures though.
I suspect this will be clearer once we can run valgrind on that branch.

Jimmy He [17:20]
Agreed on valgrind. It's also worth noting that "_if they have Eigen members_" is a big if, since most of our Eigen-related outputs are vectors of objects that contain Eigen members. They don't contain them directly. And so the Eigen::aligned_allocator gets invoked.
```

PR #1049, opened 2018-02-22
Fix "Unknown compiler version" warnings on MSVC

jdonald:
Starting with MSVC 2017 we've been spamming the logs with this:

```
Unknown compiler version - please run the configure tests and report the results
```
which comes from boost (autoboost). However, this warning is unhelpful because autoboost does not contain boost's "configure tests".

We could raise the threshold version number (1913 to be current, or higher) but that would defer to problem. Better to follow `gcc.hpp` and `clang.hpp` which do not emit warnings for untested compiler versions. When we do hit a compiler error, we'll know.

### DIFF
--- a/contrib/autoboost/autoboost/config/compiler/visualc.hpp
+++ b/contrib/autoboost/autoboost/config/compiler/visualc.hpp
@@ -288,11 +288,4 @@
 #endif
 
 //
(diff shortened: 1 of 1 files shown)


jhe-:
Ah, so that's where it came from. Yeah, no point for the compiler to speculate on potential errors.

PR #1050, opened 2018-02-23
Update standard to fix Android 32-bit compatibility

jdonald:
Upon attempting to release autowiring-1.0.6, this appears to be required to avoid breaking autowiring-configVersion.cmake compatibility for 32-bit Android.

- [ ] Figure out implications of putting this in v1.0.6 or deferring to v1.0.7


jdonald:
Discussed with @yeswalrus and although this isn't required to stabilize the current release it's better to have. It might also save us from an extra autowiring release if we happen to clean up all repos with regard to toolchain-arm32.cmake soon.

PR #1052, opened 2018-03-09
Remove support for auto_prev

gittyupagain:
The `auto_prev` feature was added to Autowiring with the intention of being able to automatically use AutoFilter decorations from one packet in the next packet. Unfortunately, we often found that there were performance penalties for doing so. It ended up being easier, and faster, to just keep a reference to the shared object if needed. Due to its issues, any references to using `auto_prev` in AutoFilters were removed from our applications.

The actual implementation of `auto_prev` relies on having successor links between the packets. This also means that any long-running packet will keep around all newer packets, even those that have completed, until the oldest remaining packet has completed.

In addition to removing the `auto_prev` feature, this PR also removes the time-shifted decoration concept, and any successor logic that links the packets together.


yeswalrus:
We'll need to bump AW to at least 1.1 after this.

PR #1055, opened 2018-03-12
Modify code to support C++17

gittyupagain:
With the release of C++17, several already deprecated STL functions have been removed. In order to support compilers that adhere strictly to the standard, this PR removes use of those old functions.

### DIFF
--- a/src/autowiring/test/ContextEnumeratorTest.cpp
+++ b/src/autowiring/test/ContextEnumeratorTest.cpp
@@ -2,6 +2,7 @@
 #include "stdafx.h"
 #include <autowiring/ContextEnumerator.h>
 #include <algorithm>
+#include <random>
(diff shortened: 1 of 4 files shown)


src/autowiring/test/ContextEnumeratorTest.cpp
```diff
@@ -146,7 +147,9 @@ TEST_F(ContextEnumeratorTest, ComplexRemovalInterference) {
     children.push_back(AutoCreateContext());
 
   // Shuffle the collection to prevent the order here from being equivalent to the order in the context
-  std::random_shuffle(children.begin(), children.end());
+  std::random_device rd;
+  std::mt19937 gen(rd());
```
yeswalrus: +1 for random_device usage

PR #1060, opened 2018-05-10
Additional fix std::filesystem with /std:c++17

### LINKED ISSUE #1059: Autowiring failed when build with MSVC
Hi all,

For the issue https://github.com/leapmotion/autowiring/issues/1057 I reported before. This issue seems not to be solved completely. For the code in filesystem.h:
```
#if !defined(_MSC_VER) || _MSC_VER < 1914 || _MSVC_LANG < 201402L
```
This is also need to change as 201403L. We are no longer hit this issue when we changed to 201403L. Could you please help take a look at this? Thanks a lot!

Thanks 
Shanshan

bellawu02:
The change in #1058 was incomplete as this constant is checked in two places. Fix it in the second one as well.

- [ ] Fixes #1059 i.e. actually fixes #1056

### DIFF
--- a/src/autowiring/C++11/filesystem.h
+++ b/src/autowiring/C++11/filesystem.h
@@ -18,7 +18,7 @@
 namespace awfsnamespace = autoboost::filesystem;
 #endif
 
-#if !defined(_MSC_VER) || _MSC_VER < 1914 || _MSVC_LANG < 201402L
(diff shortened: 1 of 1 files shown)


jdonald:
@shanshan0309 could you test this out? Sorry that it has taken too many iterations to get this right.

shanshan0309:
@jdonald. Thank you. I tried your fix and it works well.

